Issue #92, opened 2016-02-01
JSONWriter does not provide valid JSON


dneise:
We all know, that the JSONWriter does not create valid JSON files, unless `writeListOfItems = "true"` is set.
This is an old story already.

But it also does not treat infinite values and not-a-number values correctly.

We encountered files, looking like this:

```json
{"some_key" : -Infinity,
"another_key": Infinity,
"still_another_one": NaN}
``` 

As you can see, this is not valid JSON. (We also checked it using http://jsonlint.com/)

---

We use `pandas.read_json` to parse these files, and found that by exchanging these values by some string representation of infinity and nan, the parser like it much better.

E.g.
```json
{"some_key" : "-inf",
"another_key": "inf",
"still_another_one": "nan"}
```

mackaiver:
I agree somewhat. However the reasoning behind writing these values is JS compatibility.  Despite what the linter says these value are common among JavaScript parsers in many browsers. 

See also [this google statement] (https://google-gson.googlecode.com/svn/trunk/gson/docs/javadocs/com/google/gson/GsonBuilder.html#serializeSpecialFloatingPointValues%28%29) about special floating point values in JSON. 

Maybe we can add another flag for Pandas compatible output.

dneise:
Under these circumstances, JSON disqualifies as a file format for scientic data storage and exchange.

Maybe the pandas community can be asked to support the same pseudo standard as google does...

We, however, should imho stop using JSON in order not to lead our colleagues into traps.

MaxNoe:
We use `json` or `jsonl` mainly as intermediate serialisation format. Most of us convert the multiple `json`-files that come from processing multiple runs to one big file in another format. E. g. `.root` or `hdf5`.

Note that `python`'s standard library `json` module understands the `JavaScript`/`Google` Pseudo-Standard:

```{python}
In [1]: import json
In [2]: json.loads('{"a": [5, Infinity]}')
Out[3]: {'a': [5, inf]}
```

So to get a json file using the `Infinity` and `NaN` into a pandas dataframe, you can do:

```{python}
import json

data = json.load('myfile.json')
df = pd.DataFrame(data)
```

dneise:
Even worse. Using something, which is obviously not useful for scientif

mackaiver:
See also Issue #73

Issue #73, opened 2015-07-08
Output File Format


mackaiver:
Originally reported by: **Dominik Neise (Bitbucket: [dneise](http://bitbucket.org/dneise), GitHub: [dneise](http://github.com/dneise))**

---

Given that the upcoming ICRC will feature an official FACT crab spectrum, gained with fact-tools, I wonder whether it is time to create realtime analysis results on the island using fact-tools.

I understand, it is common practice in the community, to create image parameter output files, which may be used by physicists to perform their personal analysis. I understand it is not common to start off ones analysis, at the level of large raw data files, but at the level of smaller image parameter files.

I think currently, it is common practice to store these image parameters in JSON format.

It is in my opinion necessary to provide not only the results of an analysis, but also ensure reproducability and provide context. Or in other words, answer common questions like:
- what were your precuts again?
- are you sure, we used exactly the same fact-tools v0.9? 
- what forest model did you use?

I think, most questions can already be answered when having the XML and the git commit hash. (assuming that nobody would ever generate any output with a modified ans untracked fact-tools version)

I pretty much assume, there is a point of no return in this question. If we decide for a file format for these image parameters, which will turn out to be not optimal for a very common usecase, we might have to live with it.

So, what do u think? Should we stick with the current way everybody is working? Should we change something? Should we write a list of best pratices, which simply says: Make a note of what u did in the fact logbook.

What can we learn from others in the community? How is CTA going to do it? Can we do the same they will do, already today and tear down borders between experiments even before they exist?

---
- Bitbucket: https://bitbucket.org/cbockermann/fact-tools/issue/73

MaxNoe:
I think this is partly implemented and partly superceeded by our 'Eventlist' plans

PR #93, opened 2016-02-02
Refactor JSONWriter

MaxNoe:
This includes the following changes:

* `json` format by default, optional `jsonl` format by using `jsonl="true"` in the `xml`
* New `append` option, default is false, so existing files are overwritten.
* The `keys` key is now evaluated using the `stream.Keys` class, so glob patterns now work.
* Increase version number to 0.12 because these changes break xmls.

No xmls in the example folder needed to be changed, although they are no behaving differently (output is `json`, not `jsonl` now).

One might want to add an Option to convert `Infinity`, `-Infinity` and `NaN` to something strictly `json` compatible.

What would you like? `null`? Or strings like `"nan", "inf", "-inf"`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 	<artifactId>fact-tools</artifactId>
 	<name>fact-tools</name>
 
-	<version>0.11.0</version>
+	<version>0.12.0</version>
 
 	<url>http://sfb876.de/fact-tools/</url>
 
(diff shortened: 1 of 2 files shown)


MaxNoe:
See issues #92, #73

dneise:
what is `jsonl`?

MaxNoe:
http://jsonlines.org/

dneise:
Suggested Conventions

JSON Lines files may be saved with the file extension .jsonl.

MaxNoe:
The filename is choosen by the user.

dneise:
a pity ..

MaxNoe:
I added an option to serialize the special values to strings `"inf"`, `"-inf"`, `"nan"`.

I am not sure about a sensible default. Right now it is `false`, so `Infinity`, `-Infinity` and `NaN` are used by default.

What should the default be?

dneise:
I'm very glad this issue was discussed extensively here

PR #96, opened 2016-02-04
Remove obsolote processor JSONWriterExcludeKeys, fix javadocs in JSONWriter

MaxNoe:
The normal JSONWriter now supports writing all keys but a selection with
```
<fact.io.JSONWriter keys="*,!excludeKey1,!excludeKey2" url="file:test.json" />
```

This makes the `JSONWriterExcludeKeys` obsolete.


mackaiver:
Maybe we can do this and #100 in one go?

Issue #97, opened 2016-02-04
Update webpage


MaxNoe:
Our webpage still links to bitbucket:
http://sfb876.tu-dortmund.de/FACT/

MaxNoe:
Maybe we also want to additionally provide http://sfb876.tu-dortmund.de/fact-tools?

MaxNoe:
see #99

mackaiver:
I'd prefer build a new website using github pages. Maybe using the latest Jekyll. What do you think?

MaxNoe:
This will be usefull then:
https://gist.github.com/domenic/ec8b0fc8ab45f39403dd

MaxNoe:
#101

PR #99, opened 2016-02-04
update links everywhere

MaxNoe:


mackaiver:
I'm not even sure the websites being build right now. See also #97. For the old webpage some building script is running on the server hosted by @cbockermann. But I guess its pulling stuff from bitbucket still.

MaxNoe:
This will be obsolete, i will use the testpage branch for this.

PR #102, opened 2016-02-08
Processor and TypeAdapter for UnixTimeUTC to DateTime

MaxNoe:


MaxNoe:
Maybe should do this directly in our FITS Stream. One Problem is, that the DateTime Object in Java 7 only supports millisecond precision. Jaava 8 has up to nanosecond precision.

mackaiver:
Lets put this into the fitsstream. And get rid of joda time dependencies for java 8

PR #114, opened 2016-02-14
Measure performance

MaxNoe:


alexeyegorov:
isn't it something that package streams-performance should handle? or should this one has some very special "fact" purpose?

MaxNoe:
Honestly I have no idea.  The docstring says you and @mackaiver wrote this process. I just fixed it for multiple uses of the same processor

alexeyegorov:
Then, we should consider differences to streams-performance. Possibly this was written while playing around with storm before CTA meeting in September. I think we may remove this class or import some of its' special functionality (if any exists) into streams-performance.

alexeyegorov:
Actually, this is nice as it does something similar as streams-performance, but allows one to measure performance of all processors of a process container. Maybe we should think of integrating such functionality into streams-performance (@cbockermann), but at this point I think it can be merged to fact-tools.

PR #115, opened 2016-02-23
Write fits file

alexeyegorov:
- Writing FITS files using nom.tam.Fits library
- Parameters are used the same way as in JSON writer (keys, file etc.)


alexeyegorov:
some tests are failing while project can be built. This comes from refactoring branch itself as I think as I haven't changed the tests itself in here.

MaxNoe:
We now have a working fits writer in the master branch

PR #121, opened 2016-03-21
add new filter Moving Minimum

dneise:
Please advice what is missing before this new filter can be pulled.

### DIFF
--- a/src/main/java/fact/filter/MovingMinimum.java
+++ b/src/main/java/fact/filter/MovingMinimum.java
@@ -85,7 +85,7 @@ public class MovingMinimum implements StatefulProcessor{
     public void init(ProcessContext context) throws Exception {
         if((length & 1) == 0){
             length++;
-            log.warn("CentralMovingAverage only supports uneven window lengths. New length is: " + length);
+            log.warn("MovingMinimum only supports uneven window lengths. New length is: " + length);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/fact/filter/MovingMinimum.java
```diff
@@ -0,0 +1,111 @@
...
+import stream.annotations.Parameter;
+
+/**
+ * 
+ * @author Dominik Neise &lt;neised@phys.ethz.ch&gt;
+ */
+@Description(group = "Fact Tools.Filter", text = "A running mininum")
+public class MovingMinimum implements StatefulProcessor{
+
+	static Logger log = LoggerFactory.getLogger(MovingAverage.class);
```
MaxNoe: You are getting the wrong logger here. It should be `MovingMinimum.class`

src/main/java/fact/filter/MovingMinimum.java
```diff
@@ -0,0 +1,111 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import stream.Data;
+import stream.ProcessContext;
+import stream.StatefulProcessor;
+import stream.annotations.Description;
+import stream.annotations.Parameter;
+
+/**
+ *
```
mackaiver: I think this is the right place for some general information about what the processor does in a few short words.

MaxNoe:
Looks good now, any objectios @fact-project/fact-tools-core-dev ?

PR #123, opened 2016-04-01
Zfits 2012 fix

mackaiver:
This contains a number of fixes that hopefully allow for reading all FACT related .fits and zfits files. 
I tested this on a small sample of real data and monte carlo files (even data from 2012. they have weird header strings.).  Also added soem unittests.
It also cleans up some of the stuff I found in the code of  Michael B.


src/main/java/fact/io/zfits/BlockHeader.java
```diff
@@ -15,7 +15,7 @@
 		
 		private final short id;
 		// using int and casting to short saves us the work to write '(short)x' everywhere above
-		private Prozessor(int id) {
+		Prozessor(int id) {
```
MaxNoe: It's really small and nothing really important, but could we remove the german naming here?

mackaiver: Yeah. This stuff is everywhere. Didnt find all spots.

jebuss:
works for me, accepted!

jebuss:
+1

PR #132, opened 2016-05-13
Muon updates

MaxNoe:
* New Processor to calculate TimeResolution (RingStandardDeviationWithThreshold)

* Updated xml in examples/studies

* Use PixelSet for `bestRingPixel` instead of integer array

* Use camel case for default keys in the fitting classes


dneise:
+1

mackaiver:
:+1:

PR #135, opened 2016-05-19
Stop in case of error. Better error messages 

mackaiver:
The sourcepostion processor ignored all errors. I dont know whther that makes sense. I also added some (hopefully) clearer error output into the auxfileservice


ftemme:
+1

MaxNoe:
I think we can merge this

PR #136, opened 2016-07-26
Single pe extractor

relleums:
Extract single photon pulses on the pixel time lines. 
- Calculates a list of arrival time slices of photon pulses on a time line.
- Has Unit tests.

This is a port of  a python implementation were I developed the algorithm and performed some toy simulation studies.

### DIFF
--- a/examples/viewer.xml
+++ b/examples/viewer.xml
@@ -20,6 +20,11 @@
     <include url="classpath:/default/data/extraction.xml" />
     <include url="classpath:/default/data/cleaning.xml" />
 
+    <fact.extraction.SinglePulseExtraction 
+        dataKey="DataCalibrated"
+        outputKey="PhotonArrivals"
+    />
+
     <fact.ShowViewer key="DataCalibrated" />
   </process>
 </container>
--- a/examples/viewer.xml
+++ b/examples/viewer.xml
@@ -23,6 +23,7 @@
     <fact.extraction.SinglePulseExtraction 
         dataKey="DataCalibrated"
         outputKey="PhotonArrivals"
+        maxIterations="50"
     />
 
     <fact.ShowViewer key="DataCalibrated" />
--- a/src/main/java/fact/extraction/SinglePulseExtraction.java
+++ b/src/main/java/fact/extraction/SinglePulseExtraction.java
@@ -7,7 +7,6 @@ import stream.Processor;
 import stream.annotations.Parameter;
 import java.util.Arrays;
 import java.util.ArrayList;
-import fact.features.singlePulse.timeLineExtraction.ElementWise;
 import fact.features.singlePulse.timeLineExtraction.SinglePulseExtractor;
 
 /*
@@ -60,7 +59,7 @@ public class SinglePulseExtraction implements Processor {
         double[] timeLines = (double[]) input.get(dataKey);
 
         double[] single_pe_count = new double[npix];
-        ArrayList<ArrayList<Integer>> pixel_arrival_slices = 
+        ArrayList<ArrayList<Integer>> pixelArrivalSlices = 
             new ArrayList<ArrayList<Integer>>();
 
         for (int pix = 0; pix < npix; pix++) {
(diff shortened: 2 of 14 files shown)


src/main/java/fact/features/singlePulse/timeLineExtraction/AddFirstArrayToSecondArray.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class AddFirstArrayToSecondArray {
+    
+    public static void at(double[] f1, double[] f2, int at) {
+        if(at > f2.length) return;
+        if(at < 0) at = 0;
+        // endpoint of injection in f2: e2
+        int e2 = at + f1.length;
+        // naive endpoint of sampling in f1
+        int e1 = f1.length;
```
mackaiver: I don't think this variable is ever used. Can it be removed?

mackaiver:
Looks nice. I removed some unused imports and stuff. Do you want me to push these changes here? Otherwise I create another pull request.

src/main/java/fact/features/singlePulse/timeLineExtraction/AddFirstArrayToSecondArray.java
relleums: This turned out to be a bug. I tracked it down with two more test cases.
Looks like I missed sth. from the python port.

relleums:
Good catch sir. I think this is because I do not use any IDE.
I will work over the branch once more myself and try to make the comment blocks more like you make them look.

src/main/java/fact/features/singlePulse/timeLineExtraction/AddFirstArrayToSecondArray.java
```diff
@@ -0,0 +1,57 @@
...
+        int end = at + first.length;
+        if(end < 0) return;
+            // first array is completley before second. No overlap.
+            //    [...first...]
+            //                   [...second...]
+            //     ^at
+
+        if(end >= second.length) end = second.length;
+            // end is limited to length of second
+            // first array is completley before second. No overlap.
```
MaxNoe: copy paste error?

MaxNoe:
I just threw it on the test file, there is a nice correlation between the number of found photons and the photoncharge. However, this method seems to tend to find more photons than the photoncharge..

photoncounts

relleums:
The correlation looks actually better than I expected it to be, given that the 'fine tuning' of the single pe extractor was only done in my python MC toy simulation so far. Anyhow, I did not longer want to keep the single pe extractor to myself. We can now together investigate the fine tuning on observed data.
We need a tool in fact-tools where we can simulate time lines with pulses on them. This way we can compare all our extractors VS. the MC truth. So photoncharge vs MC truth and photonArrivalsCounts vs MC truth.

MaxNoe:
Yes. Really nice work.

I would opt for merging.

Could you maybe also provide a small discription in the CHANGELOG.md and increment the version number?

dneise:
Yes, the single pe extractor finds also singles outside the integration window. So the sum of found singles must always be larger or equal than the photon charge.

relleums:
I just saw that I forgot to copy an important step of my python sandbox (normalizing the convolution kernel). Now the performance for small pulses (1-5) is more accurate. I also simplified the interface and added unit tests for the overall SinglePulseExtractor.

mackaiver:
Ready to merge right?

MaxNoe:
LGTM

examples/viewer.xml
```diff
@@ -20,6 +20,12 @@
     <include url="classpath:/default/data/extraction.xml" />
     <include url="classpath:/default/data/cleaning.xml" />
 
+    <fact.extraction.SinglePulseExtraction 
+        dataKey="DataCalibrated"
+        outputKey="PhotonArrivals"
+        maxIterations="50"
+    />
```
MaxNoe: I would take this out here. Viewer should kind of be the minimal example.

src/main/java/fact/extraction/SinglePulseExtraction.java
```diff
@@ -0,0 +1,152 @@
...
+    private void printArrivalSlices(ArrayList<ArrayList<Integer>> arr) {
+        for(int pix=0; pix<arr.size(); pix++) {
+            System.out.print("pix "+pix+": ");
+            for(int ph=0; ph<arr.get(pix).size(); ph++)
+                System.out.print(arr.get(pix).get(ph)+" ");
+            System.out.print("\n");
+        }
+    }
+
+    public String getDataKey() {
```
MaxNoe: We do not need getters. They just ad clutter.

src/main/java/fact/extraction/SinglePulseExtraction.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public Data process(Data input) {
+
+        Utils.isKeyValid(input, "NPIX", Integer.class);
+        Utils.mapContainsKeys(input, dataKey,  "NPIX");
+
+        npix = (Integer) input.get("NPIX");
+        roi = (Integer) input.get("NROI");
+        double[] timeLines = (double[]) input.get(dataKey);
```
jebuss: To be consistent with other processors I suggest to call this array "data"

src/main/java/fact/features/singlePulse/timeLineExtraction/ArgMax.java
```diff
@@ -0,0 +1,27 @@
...
+    public double max;
+
+    /**
+    * @param arr
+    *           The array to find its maximum value and the maximum value's
+    *           position of.
+    */
+    public ArgMax(double[] arr) {
+        arg = 0;
+        max = arr[0];;
```
jebuss: double ";"

src/main/java/fact/features/singlePulse/timeLineExtraction/ArgMax.java
```diff
@@ -0,0 +1,27 @@
+package fact.features.singlePulse.timeLineExtraction;
+
+import java.util.Arrays;
+/**
+* Finds the maximum on an array and stores its position (arg) 
+* and its value (max).
+*/
+public class ArgMax {
```
jebuss: shouldnt it be ArrMax or ArrayMax?

src/main/java/fact/features/singlePulse/timeLineExtraction/TemplatePulse.java
```diff
@@ -0,0 +1,66 @@
...
+
+        final double periodeSliceInNs = 0.5;
+
+        final double[] time = new double[lengthInSlices];
+        for (int i = 0; i < time.length; i++) {time[i] = i*periodeSliceInNs;}
+
+        double[] template = new double[lengthInSlices];
+        for (int i = 0; i < time.length; i++) {
+
+            final double amplitude =
```
jebuss: Maybe add a comment with the source of this parameterization?

jebuss:
LGTM2

MaxNoe:
The `ArgMax` class does not return anything. Maybe a static method returning something would be better.

PR #138, opened 2016-08-09
Fits writer master

MaxNoe:
I cherry picked @alexeyegorov commits to have the changes in the master, not the refactoring branch

I added a few minor changes, which make it possible to read in the written file with fact tools again.

This can now write fits files and read the result. One could write out photoncharges/arrivaltimes and do the analysis later.


MaxNoe:
This FITS writer has some limitations:

* It keeps all the stuff in memory until the end of the stream (write operation happens in `finish`)
* It does not support variable length columns

MaxNoe:
I added a new `Processor` called `StreamingFITSWriter` which writes out each data item directly and does not keep the whole table in RAM.

It sadly also does not support variable length arrays like for example `shower` pixels.

MaxNoe:
What do you think: Should we keep both Processors or just the `StreamedFITSWriter`? Possibly as `FITSWriter`

MaxNoe:
Also I do not like how all those type checks happen twice. Do have an idea how to do this better?

MaxNoe:
Reading in the fits file using astropy ist ~50 times faster than using `json.load` on a json file containing the same data.

MaxNoe:
@fact-project/fact-tools-core-dev 

I think this is ready for review now. Any comments?

src/main/java/fact/io/FITSWriter.java
```diff
@@ -0,0 +1,254 @@
...
+ *
+ * The fits file is initialised using the given keys from first data item.
+ * All following items must have the same structure.
+ *
+ */
+public class FITSWriter implements StatefulProcessor {
+
+    static Logger log = LoggerFactory.getLogger(FITSWriter.class);
+
+    @Parameter(required = true)
```
jebuss: description?

src/main/java/fact/io/FITSWriter.java
```diff
@@ -0,0 +1,254 @@
...
+
+    public void setKeys(Keys keys) {
+        this.keys = keys;
+    }
+
+    public void setUrl(URL url) {
+        this.url = url;
+    }
+
+    public void setExtname(String extname) {
```
jebuss: setExtName

jebuss:
LGTM

PR #140, opened 2016-08-15
Jebuss/neighbor pix dcf

jebuss:
* NEW FEATURE: NeighborPixelDCF calculates the Descrete Correlation Function for neighboring pixel time series and returns the moments of their distribution over the neighbor pixels
* renamed MeanCorrelation to NeighborPixelCorrelation and added more statistical moments as feature

### DIFF
--- a/src/main/java/fact/Utils.java
+++ b/src/main/java/fact/Utils.java
@@ -7,12 +7,14 @@ import com.google.common.primitives.Ints;
 import fact.container.PixelSet;
 import fact.hexmap.FactCameraPixel;
 import fact.hexmap.FactPixelMapping;
+import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
 import org.slf4j.Logger;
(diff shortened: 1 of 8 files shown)


src/main/java/fact/Utils.java
```diff
@@ -468,7 +470,56 @@ public static void checkWindow(int start, int size, int validLeft, int validRigh
...
 
 
+	/**
+	 * Absolute position in the data array by pixel and slice
+	 *
+	 * @param pix
+	 * @param slice
+	 * @return absolute array position
+	 */
+	public static int absPos(int pix, int slice, int roi) {
```
MaxNoe: We need this all over the place. Maybe put it in Utils?

MaxNoe:
I think we can merge this, #138 and #136 and make that `0.16.0`. What do you guys think? @fact-project/fact-tools-core-dev

jebuss:
Agrrrrreeeee!

PR #141, opened 2016-08-16
Timelapse calibration

MaxNoe:
This adds Processors to be able to do DRS4 Calibration like MAGIC does it.

Code to produce the calibration fits file is here: https://github.com/fact-project/timelapse_calibration

A file calculated with the run `20160728_018.fits.fz` is attached. Note that a few fits failed in this file.
We need to take more data to get better results.

### DIFF
--- a/examples/timelapse_data_extraction.xml
+++ b/examples/timelapse_data_extraction.xml
@@ -1,0 +1,16 @@
+<container>
+  <properties url="classpath:/default/settings.properties" />
+  <!-- Pathes to the input files -->
+  <property name="infile" value="file:src/main/resources/testDataFile.fits.gz" />
+  <property name="outfile" value="file:timelapse_data.json" />
+
+  <stream id="fact" class="fact.io.zfits.ZFitsStream"  url="${infile}" limit="50"/>
+
+  <!-- Description of the process and the corresponding stream -->
+  <process id="2" input="fact">
+    <fact.io.PrintKeysOnConsole keys="EventNum" />
+    <fact.utils.BoardTimeToSeconds outputKey="eventTime"/>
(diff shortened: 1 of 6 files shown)


MaxNoe:
Some plots to illustrate:

## General Idea

In drs4 chips, there is a clear dependency between the time since a cell was last read out and its offset value:
fit_result_p0_c0

This can be parameterized as `offset = a * delta_t ^ b + c`. In the plot you can see the fit result for Pixel 0, Cell 0.

## Results

Here are some FACT-Tools Viewer results comparing `DrsTimelapseCalibration` calibrated data to our traditional `DrsCalibration`. Note that tha calibration file is only valid for the period after the FAD Board was changed. You can see the effect in the lower left of the camera if you look into the standard test file.

calibration_result

dneise:
I think one should somehow stress the fact, that this TimelapseCalibration in a sense indends to make the jump correction obsolete.

So it might be interesting to show an event containing a nicely visible jump with
 * std drs amplitude calib --> jump visible
 * after jump correction --> no jump
 * alternatively after Timelapse --> also no jump(?)

MaxNoe:
Here an example:

jump

Blue TimelapseCalibration, Cyan conventional DrsCalibration, Red Conventional + JumpCorrection

MaxNoe:
jump2

MaxNoe:
jump3

dneise:
Something seems wrong, can you confirm the name - color relation again?

MaxNoe:
Yes, colors are correct. You can also see the labels (DataRemoved = After JumpRemoval)

MaxNoe:
This is to WIP

PR #144, opened 2016-08-31
Delete some potentially unused stuff

mackaiver:
Hello!
I want to start a discussion about potentially unused processors and other code.
Can we remove these? Which one should we keep? Which ones can be removed?


jebuss:
Yeah, Discussion ;-) !!!

We should not do 4d8e6b8 and delete src/main/java/fact/features/singlePulse/PulseMaxAmplitude.java.
It is needed for the single pulse extraction do single PE spectra. The whole process for this is not finished. But I agree we need documentation and we need an example process.

src/main/java/fact/features/video/CenterOfGravity.java
```diff
@@ -1,301 +0,0 @@
...
-{
-    
-    FactPixelMapping pixelMap = FactPixelMapping.getInstance();
-    
-    private int sliceCount;
-    
-    /**
-     * This function calculates the center of gravity for every slice. It uses only shower pixel.
-     * It also calculates the variance and covariance of the center of gravity.
-     */
```
jebuss: This processor was developed by Jan Freiwald. You

mackaiver: I know. Do we need it? It basically 'tracks the position of the cog' right? Maybe that can be a nice feature for G/H separation. But right now we don't use it right?

jebuss: yes, we should have a look in his thesis if it was promissing, check its runtime and then decide to add it the stdAnalysis or delete it.

PR #145, opened 2016-08-31
Fix javadocs and some imports in sebastians single pe extractor

mackaiver:
The title says it all. This is basically cosmetics. I however would like to discuss how to handle methods that modify an array in-place instead of returning it. Personally I prefer to return the modified object. Or maybe even better yet: copy the thing and return a new thing. What do you think?

### DIFF
--- a/src/main/java/fact/features/singlePulse/timeLineExtraction/AddFirstArrayToSecondArray.java
+++ b/src/main/java/fact/features/singlePulse/timeLineExtraction/AddFirstArrayToSecondArray.java
@@ -1,27 +1,31 @@
 package fact.features.singlePulse.timeLineExtraction;
 
-import java.util.Arrays;
-
+/**
+ * Utility class for adding elements of two arrays
+ */
 public class AddFirstArrayToSecondArray {
 
     /**
-    * Adds the first array elementwise to the second array starting at a certain
-    * position. 
-    *
-    * @param first
-    *           The first array [N].
-    *
-    * @param second
-    *           The second array [M] is modified in place. The length M of the 
-    *           second array will not be changed.
-    *
-    * @param at
-    *           The starting position of the addition in the second array. 
-    *           Can be negative.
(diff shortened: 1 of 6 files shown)


MaxNoe:
I would also prefer to return a copy.

relleums:
I also vote for returning a copy, as it sounds like "least surprise".  I think 

```
public static void applyAcCoupling(double[] timeLine) {

        if(timeLine.length == 0)
            return;

        double sum = 0.0;
        for (double aTimeLine : timeLine){
            sum += aTimeLine;
        }

        final double mean = sum/(double)(timeLine.length);

        timeLine = ElementWise.add(timeLine, -mean);
}
```

is such a candidate

relleums:
The element used in the for loop is not a timeLine anymore, it is a slice (I would call it a slice).
`        for (double aTimeLine : timeLine){
            sum += aTimeLine;
}`
Is the additional letter 'a' a common thing?
Also here
`for (double aPulseToLookFor : pulseToLookFor){
            sum += aPulseToLookFor;
}`
I would call it a slice instead of aPulseToLookFor.

MaxNoe:
I never saw an article in a variable name. I would go for having a either a plural form for the container variable or the addition of the container.

mackaiver:
> Is the additional letter 'a' a common thing?
Nope. I just messed up ^_^

relleums:
I vote for

- not fixing the 'return copy' issue yet
- either using my old explicit for loop with the running index or using the variable name 'slice'  or 'timeSlice' instead of 'aTimeLine' for your range based for loop.
- fixing all the javadocs and imports, as it is implied by your branch name

and then, merging back.

Jens and I already now know that we want to change/finalize the single pulse extractor API before going in production (returning arrays of arrays instead of ArrayLists of ArrayLists). I would like to branch and work in these changes the moment you got the imports and javadocs corrected.

PR #148, opened 2016-09-13
Ignore borked ceres files. 

mackaiver:
This is a suggestion (you might call it a dirty hack) to deal with 'empty' ceres files. 
Sometimes ceres writes fits.gz files with a file size of < 5KB.  These contain part of a .fits header. However the `END` keyword is missing. thus they are not valid fits files imho. Trying to parse these files with the zfits reader fails. This pull request just checks the file size. If its less than 5 KB it will not try to parse it.   

Maybe we should not merge this but fix ceres instead...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <groupId>de.sfb876</groupId>
   <artifactId>fact-tools</artifactId>
   <name>fact-tools</name>
-  <version>0.16.0</version>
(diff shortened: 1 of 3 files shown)


mackaiver:
I don't really like this since its a dirty hack. In my opinion the reader should fail when it encounters a broken fits file. This should be fixed in ceres.

MaxNoe:
bork

PR #149, opened 2016-09-25
Fix another mixup between rows and columns

mackaiver:
Also added a test file which @dneise provided here https://github.com/esheldon/fitsio/issues/94


MaxNoe:
As this is rather urgent, I would call it 0.16.2 and merge. @fact-project/fact-tools-core-dev  any objections?

jebuss:
LGTM

Issue #71, opened 2015-06-23
Aux Data not read because of String Index Out of bounds


mackaiver:
Originally reported by: **Maximilian Nöthe (Bitbucket: [MaxNoe](http://bitbucket.org/MaxNoe), GitHub: [MaxNoe](http://github.com/MaxNoe))**

----------------------------------------

Analysing the data between October 2013 and May 2014 I had several
days of datataking where facttols failed to load the aux service with the following error:

```
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1911)
	at fact.auxservice.AuxFileService$1.accept(AuxFileService.java:119)
	at java.io.File.list(File.java:1087)
	at fact.auxservice.AuxFileService.findAuxFileUrls(AuxFileService.java:112)
	at fact.auxservice.AuxFileService.getAuxiliaryData(AuxFileService.java:49)
	at fact.features.source.SourcePosition.process(SourcePosition.java:206)
	at stream.runtime.AbstractProcess.process(AbstractProcess.java:126)
	at stream.runtime.AbstractProcess.execute(AbstractProcess.java:186)
	at stream.runtime.ProcessThread.run(ProcessThread.java:125)
```

The affected dates are:

* 2013-10-26
* 2013-11-21
* 2013-11-23
* 2014-04-04
* 2014-04-17

----------------------------------------
- Bitbucket: https://bitbucket.org/cbockermann/fact-tools/issue/71

Issue #152, opened 2016-10-17
huge output to stdout, during "std-analysis"


dneise:
When doing some "std-data-analysis" runs at ISDC, this is a "most typical" std out of fact-tools.

Can one remove a bit of this? What about these?
- error due to something with hdfs.
- warnings of `AuxFileService` regarding files it does not know...

Or maybe one can think about this the other way round. What is the output a user might want to have?
Basically none I think. A progressbar would be awesome for single jobs on user machines, but is useless on clusters. The total runtime, one measures using `time`. The file which is currently worked on, one should know or `echo` it in case one does not. 

```
java -jar -Xmx2500M -Xms2500M     /home/guest/neise/facttools_massive_production/single_photon_extractor/fact-tools-v0.16.0.jar     /home/guest/neise/facttools_massive_production/single_photo
n_extractor/analysis.xml     -DauxFolder=file:/fact/aux/2014/05/25/     -Doutfile=file:/scratch/fact/single_photon_extractor/json/2014/05/25/20140525_170.json.gz     -Dinfile=file:/fact/raw/2
014/05/25/20140525_170.fits.fz     -Ddrsfile=file:/fact/raw/2014/05/25/20140525_145.drs.fits.gz
Oct 17 2016 09:10:47 [ERROR] stream.io.SourceURL: Failed to register handler for protocol 'hdfs'
Oct 17 2016 09:10:47 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/prevEventAndSkip.xml'
Oct 17 2016 09:10:47 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/calibration.xml'
Oct 17 2016 09:10:47 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/extraction.xml'
Oct 17 2016 09:10:47 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/cleaning.xml'
Oct 17 2016 09:10:47 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/parameterCalc.xml'
Oct 17 2016 09:10:47 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/sourceParameter.xml'
Oct 17 2016 09:10:48 [INFO ] stream.runtime.ProcessContainer: ######## Sources ########
[...]

Issue #151, opened 2016-10-17
"Aborting process due to errors: String index out of range: -1"


dneise:
While doing some raw data analysis on the FACT data cluster at ISDC, I encountered this error.

I would assume it might have something to do with the auxfiles ... not sure though.
An exception text, which clearly states what the error is, would be good.

```
java -jar -Xmx2500M -Xms2500M     /home/guest/neise/facttools_massive_production/single_photon_extractor/fact-tools-v0.16.0.jar     /home/guest/neise/facttools_massive_production/single_photon_extractor/analysis.xml     -DauxFolder=file:/fact/aux/2014/04/17/     -Doutfile=file:/scratch/fact/single_photon_extractor/json/2014/04/17/20140417_187.json     -Dinfile=file:/fact/raw/2014/04/17/20140417_187.fits.fz     -Ddrsfile=file:/fact/raw/2014/04/17/20140417_168.drs.fits.gz
Oct 14 2016 01:10:02 [ERROR] stream.io.SourceURL: Failed to register handler for protocol 'hdfs'
Oct 14 2016 01:10:02 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/prevEventAndSkip.xml'
Oct 14 2016 01:10:02 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/calibration.xml'
Oct 14 2016 01:10:02 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/extraction.xml'
Oct 14 2016 01:10:02 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/cleaning.xml'
Oct 14 2016 01:10:02 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/parameterCalc.xml'
Oct 14 2016 01:10:02 [INFO ] stream.util.XIncluder:    url expanded to: 'classpath:/default/data/sourceParameter.xml'
Oct 14 2016 01:10:05 [INFO ] stream.runtime.ProcessContainer: ######## Sources ########
Oct 14 2016 01:10:05 [INFO ] stream.runtime.ProcessContainer: ########fact.io.zfits.ZFitsStream@bcbe4b0########
Oct 14 2016 01:10:05 [INFO ] stream.runtime.ProcessContainer:   ==> DefaultProcess['1']
[...]

PR #153, opened 2016-10-17
Improve on reported issues 71, 151, 152

mackaiver:
This fixes the problem described in #71. It appeared because I was to stupid to catch the right exception.


mackaiver:
Oh yeah. And also #151

mackaiver:
Oh and it also reduces the standard output greatly. as mentioned in #152

dneise:
awesome

Issue #162, opened 2016-11-15
singlePulseExtractionImprovement: Define maxIterations


jebuss:
Right now the maximum number of iterations is handover from the XML. For some events it might be far too many, for others far to less. I suggest to correlate this number with the maxAmplitude in the time window which is neglecting the time marker pulse.

I thinks we should not make this number a free parameter.

PR #164, opened 2016-11-21
Single pulse extraction improvement

relleums:
The API is much more flexible now (no more static global variables). I would like to make this a production release (in the master) as soon as possible to process all our data at ISDC before it is only available on tapes.


src/main/java/fact/Utils.java
```diff
@@ -522,4 +523,17 @@ public static int absPos(int pix, int slice, int roi) {
...
 	}
 
+
+	/**
+	 * Flatten a 2d array
+     *
+     * @param array2d 2dim double array
+     * @return flattend double array
+	 */
+	public static double[] flatten2dArray(double[][] array2d) {
```
mackaiver: No sure we need an extra function for this.  The one line you wrote here is pretty nice and already right?  Maybe it can be used directly.

src/main/java/fact/features/singlePulse/timeSeriesExtraction/SinglePulseExtractor.java
```diff
@@ -0,0 +1,164 @@
...
+        public double timeSeriesBaseLine() {
+            DescriptiveStatistics statistics = 
+                new DescriptiveStatistics(timeSeriesAfterExtraction);
+            return statistics.getMean();
+        }    
+    }
+
+    public Config config;
+
+    public double[] pulseToLookFor;
```
mackaiver: I'm not sure, but it looks to me like some of these variables could be made final. Just to add a bit more rigor to the class. So nobody gets the idea to change these references.

src/main/java/fact/utils/ConvertSinglePulses2Timeseries.java
```diff
@@ -0,0 +1,90 @@
...
+import fact.features.singlePulse.timeSeriesExtraction.SinglePulseExtractor;
+import fact.features.singlePulse.timeSeriesExtraction.TemplatePulse;
+import fact.features.singlePulse.timeSeriesExtraction.ElementWise;
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.math3.analysis.function.Sin;
+import stream.Data;
+import stream.Processor;
+import stream.annotations.Parameter;
+
+import javax.rmi.CORBA.Util;
```
mackaiver: I don't know what this is. Probably an accidental import right? sometimes intellij is too fast for me ^^

jebuss:
I also think that this can be merged. If it is not superduper urgent I would suggest to have look on friday and then merge it. Maybe we could even fix #162 by then.

src/main/java/fact/Utils.java
jebuss: We decided to keep this for now.

src/main/java/fact/utils/ConvertSinglePulses2Timeseries.java
jebuss: done

jebuss:
We still have to write a changelog

PR #165, opened 2016-11-21
Update auxfileservice

mackaiver:
With this update you can simply specify the path to a folder containing all the auxiliary files in its subfolders. Like this basically:

```
<service id="auxService" class="fact.auxservice.AuxFileService" auxFolder="file:/fact/aux/2015" />
.
.
.
<fact.features.source.SourcePosition    outputKey="sourcePosition"    auxService="auxService" />
``` 

Also, and this might be important, currently this thing updates its file list every 12 hours. I hope this is useful for long running fact-tools processes like the RTA. In other words: This will scan the whole filesystem below the path you specified.


mackaiver:
Hello. Someone willing to test and review this? Would be awesome.

MaxNoe:
I don't really get why it is necessary to parse the whole filesystem below `auxFolder`.
We have a `YYYY/MM/DD` structure everywhere we have those files (or could simply create it for e.g. unit tests).

Then one could simply look up the file only for the requested timestamp.

mackaiver:
Look at this now. It just checks for the file. Needs to be in the canonical file structure. Added some minor things as well.

MaxNoe:
I fucked up with git, good version is #200

Issue #161, opened 2016-11-15
minimal viewer example and SinglePulseExtractor


relleums:
In the first SinglePulseExtractor (SPE) implementation which came back into the master in 6d33631331b0258973a897bc699ab996e807b545, I added the SPE processor to the ```/examples/viewer.xml```. Now Max N. argued, that the viewer.xml shall be minimal and not be filled up with each new feature. So far so good. BUT what is then the ```/examples/viewer_minimal.xml``` about? And why has the ```viewer_minimal.xml``` even more loc then the ```viewer.xml```?
Shall I remove the ```viewer_minimal.xml``` and remove the SPE processor from the ```viewer.xml``` and maybe make a dedicated ```viewer_single_pulse_extraction.xml```?

MaxNoe:
The viewer_minimal has more locs but no containers loaded. I think that was the point of the `viewer_minimal.xml`, not hiding stuff in containers.

PR #166, opened 2016-11-25
deleted minimal_viewer.xml

mackaiver:
I think this file can be removed right?


relleums:
confirmed, this also caused confusion in issue #161

PR #170, opened 2016-12-05
Additional pedestal parameters for std analysis

jebuss:
Added calculation of pedestal parameters e.g. ped_var to the std. Analysis
Raise version to 0.17.2


MaxNoe:
If we are now changing this stdAnalysis xml anyway, could you maybe clean all funny pathes (`/home/fabian`) up and remove the ROOTAsciiWriter (substituting propably with the FITSWriter).

PR #177, opened 2016-12-14
A new Fits Reader

mackaiver:
This is a new fits reader implementation. It can read .fits,  .fits.gz,  .fits.fz. At least on all files I tested. I tested some random files from 2013, 2014, 2015, 2016. 

There is documentation in the Java docs. Some zfits format details should mabye be documented somewhere else as well? I don't know.

Some of the unit tests compare the new reader to the old implementations. 

I'm thinking of putting this into an external module and then importing it into the FACT-Tools with maven. The only dependency for this module is Guava. There is no direct dependency to streams. Not sure if that makes sense. 

Anyways here's wonderwall:

```
/**
 *
 * A Fits object containing all HDUs in a given file.
 * HDUs can be accessed by their name.
 *
 *    Fits f = Fits.fromPath(p)
 *    HDU events = f.getHDU("ZDrsCellOffsets");
 *    BinTable binTable = events.getBinTable();
 *
 * BinTables can also directly be accessed by name if they exist.
 *
 *    Fits.fromPath(p).getBinTableByName("Events")
 *                    .ifPresent(binTable -> {
 *                        //do something with bintable
 *                    })
 *
 * Data from the BinTable can be read using the BinTableReader and ZFitsHeapReader classes.
 *
 *    Fits f = new Fits(u);
 *    BinTable events = f.getBinTableByName("Events").orElseThrow(IOException::new);
 *
 *    for(OptionalTypesMap<String, Serializable> p : BinTableReader.forBinTable(events)){
 *      assertTrue(p.containsKey("Data"));
 *    }
 *
 *
 * Created by mackaiver on 03/11/16.
 */
```


src/main/java/fact/io/hdureader/BinTable.java
```diff
@@ -0,0 +1,256 @@
...
+
+
+    /**
+     * This enum maps the type characters in the header to the fits types.
+     * Not all types from the fits standard are supported here.
+     *
+     * Note: support variable length arrays.
+     *
+     */
+    enum ColumnType {
```
MaxNoe: What are the members of these enums?

src/main/java/fact/io/hdureader/ZFitsHeapReader.java
```diff
@@ -0,0 +1,507 @@
...
+        for (int i = 2; i < data.length; i++) {
+            data[i] = (short) (data[i] + (data[i-1] +  data[i-2])/2 );
+        }
+    }
+
+    /**
+     * This methods starts by building a huffman tree from the symbol table in the data.
+     * Once build it will be used to map code words to the symbols.
+     * Here goes the code from huffman.h in FACT++ which describes how the symbol table is written to the data.
+     * apparently this is the only documentation about that.
```
MaxNoe: Maybe add quotes around `documentation`?

MaxNoe:
I think the `OptionalTypesMap` should be iterable in a foreach loop like this:

```
for(Map.Entry<String, Object> entry: row){
    log.info(entry.getKey());
}
```

mackaiver:
Add some more context please.  What file did you use? The statement `row.get("Data");` in the unit tests works just fine.

mackaiver:
The ZSHRINK keyword is not used in this implementation because the catalog is not used either. This is simply streaming the binary blob in the heap of the bintable. hence no random access.

src/main/java/fact/io/hdureader/BinTable.java
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Representation of column meta data for a bintable.
+     * Keeps information like the repeatcount, type and name of the column.
+     */
+    class TableColumn{
+
+        int repeatCount = 0;
+        ColumnType type;
+        String name;
+
```
dneise: Fits tables can have many more parameters, in FACT one often finds at least the "UNIT" parameter.
It might be interesting for people to read this, even if it is often not used according to the Fits standard.

src/main/java/fact/io/hdureader/BinTable.java
```diff
@@ -0,0 +1,256 @@
...
+     *  1. The TFIELDS keyword must exist.
+     *  2. The number of columns stored in the TFIELD line has to fit the number
+     *     of TTPYE fields
+     *  3. Make sure the NAXIS2 keyword exists. It gives the number of rows.
+     *  4. NAXIS2 needs to be non-negative.
+     *
+     * @param header the header to check
+     */
+    private void binTableSanityCheck(Header header) {
+        int tfields = header.getInt("TFIELDS").orElseThrow(() -> {
```
dneise: `tfields` here is equal to (and has to be equal to) `numberOfColumnsInTable = columns.size();`
further above. I have the feeling some stuff is done in a duplicate way here. This might be a little confusing to the reader...

src/main/java/fact/io/hdureader/BinTableReader.java
```diff
@@ -0,0 +1,149 @@
...
+
+        public OptionalTypesMap<String, Serializable> getNextRow() throws IOException {
+            OptionalTypesMap<String, Serializable> map = new OptionalTypesMap<>();
+            for(BinTable.TableColumn c : columns){
+                if(c.repeatCount > 1){
+                    map.put(c.name, readArrayFromStream(c, stream));
+                } else if(c.repeatCount == 1) {
+                    map.put(c.name, readSingleValueFromStream(c, stream));
+                } else if(c.repeatCount == 0){
+                    map.put(c.name, null);
```
dneise: is `null` the right thing to return here? can it make things break for people like:
Just read the file plain in with facttools and write it to a jsonl; will this work with this `null`? If yes, its cool.

src/main/java/fact/io/hdureader/Fits.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private final URL url;
+    private final List<HDU>  hdus = new ArrayList<>();
+
+    private Map<String, HDU> hduNames = new HashMap<>();
+
+    public HDU primaryHDU;
+
+    /**
+     * Creates a fits instance from a path object without throwing checked exceptions immediately.
```
dneise: I don't understand this `without throwing checked exceptions immediately.` part what are *checked* exceptions?

src/main/java/fact/io/hdureader/Fits.java
```diff
@@ -0,0 +1,194 @@
...
+
+        } catch (EOFException e){
+            primaryHDU = hdus.get(0);
+            stream.close();
+
+            log.debug("A total of {} HDUs were found in the file.", hdus.size());
+        }
+    }
+
+    static DataInputStream getDecompressedDataStream(URL url) throws IOException {
```
dneise: this name `getDecompressedDataStream` when I first read it somewhere further up ... I was wondering how you were able to decompresse the zfits data already that early in the process. Now I see that `decompress` refers to gzip only. Maybe renaming the function in a way, that makes clear this is about gzip only and not about our own zfits stuff might make sense.

I have no idea for a good name though.

src/main/java/fact/io/hdureader/Fits.java
```diff
@@ -0,0 +1,194 @@
...
+            if(h.equals(hdu)){
+                bytesToSkip += h.header.headerSizeInBytes;
+                break;
+            }
+            bytesToSkip += h.header.headerSizeInBytes + h.sizeOfDataArea();
+        }
+
+
+        long skipped = stream.skip(bytesToSkip);
+        if(bytesToSkip!= skipped ){
```
dneise: This is now the 3rd time I see this check for the number of skipped bytes. 
Is it possible to enhance `stream.skip()` to a `stream.skip_raises()` which encapsulates this check for you, so you can't accidentally forget it?

src/main/java/fact/io/hdureader/BinTableReader.java
mackaiver: We had some debate about this.
My first instinct was to not add it at all. Not sure if thats wise however.
Putting null at this point might not be too bad though. What the users sees in the end is the OptionalTypesMap which has getter methods for returning Optionals. At this point I hope NPEs are avoided.

src/main/java/fact/io/hdureader/Fits.java
mackaiver: There is some debate whether checked exceptions make sense at all. Many languages don't have them. C# and Python to name two. I'm a friend of unchecked exceptions. But thats just my personal opinion.  There is a whole lot of discussion about the issue

http://www.yegor256.com/2015/07/28/checked-vs-unchecked-exceptions.html

https://www.ibm.com/developerworks/library/j-jtp05254/

src/main/java/fact/io/hdureader/Fits.java
mackaiver: It might be yes. I'm a bit reluctant to wrap the stream in another layer. Not sure if that actually improves readability. I could add another functions which performs this check. But you know. Then its another method with a side effect.

src/main/java/fact/io/hdureader/BinTable.java
```diff
@@ -0,0 +1,256 @@
...
+        String typeString;
+        Class typeClass;
+        int byteSize;
+
+        ColumnType(String typeString, Class typeClass, int byteSize) {
+            this.typeString = typeString;
+            this.typeClass = typeClass;
+            this.byteSize = byteSize;
+        }
+        public static ColumnType typeForChar(String type){
```
mackaiver: It kind of behaves like an immutable map. However enums have to be completely defined at compile time. allowing for such things as switch cases or more explicit conditional statements as in `if (type = ColumnType.SHORT)` See ZFitsHeapReader line 464 and following.

src/main/java/fact/io/hdureader/ZFitsHeapReader.java
mackaiver: Done

src/main/java/fact/io/hdureader/BinTable.java
mackaiver: Done

src/main/java/fact/io/hdureader/FitsStream.java
```diff
@@ -0,0 +1,110 @@
...
+    public FitsStream(SourceURL url){
+        this.url = url;
+    }
+
+    @Override
+    public void init() throws Exception {
+        super.init();
+
+        //create a fits object
+        URL url = new URL(this.url.getProtocol(), this.url.getHost(), this.url.getPort(), this.url.getFile());
```
mackaiver: Done

src/main/java/fact/io/hdureader/Fits.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.*;
+import java.util.zip.GZIPInputStream;
+
+/**
+ *
+ * A Fits object containing all HDUs in a given file.
+ * HDUs can be accessed by their name.
+ *
+ *    Fits f = Fits.fromPath(p)
+ *    HDU events = f.getHDU("ZDrsCellOffsets");
```
mackaiver: Fixed

src/main/java/fact/io/hdureader/BinTable.java
```diff
@@ -0,0 +1,256 @@
...
+
+        int repeatCount = 0;
+        ColumnType type;
+        String name;
+
+        /**
+         * Create a TableColumn from the tform and ttype header information.
+         * @param tform the headerline specifying the type of the data.
+         * @param ttype the headerline specifying the name of the column.
+         */
```
mackaiver: Fixed

src/main/java/fact/io/hdureader/BinTable.java
```diff
@@ -0,0 +1,256 @@
...
+        numberOfColumnsInTable = columns.size();
+
+
+        // create readers for the data in this table
+        //create stream for data in table area of the bintable extension
+        DataInputStream stream = Fits.getDecompressedDataStream(url);
+
+        long skippedBytes = stream.skipBytes((int) (hduOffset + header.headerSizeInBytes));
+        if(skippedBytes != hduOffset + header.headerSizeInBytes){
+            throw new IOException("Could not skip all bytes to table data in this HDU.");
```
mackaiver: Added a better message

src/main/java/fact/io/hdureader/BinTable.java
mackaiver: Removed some duplication

mackaiver:
The Max found is now fixed.

mackaiver:
> Also: why do we keep the FitsStream and the zfits.ZFitsStream. I think we should get rid of them.

I would deprecate them for now. Taking them out will introduce some changes for services and processors that read fits files. We should do it as soon as possible but not in this PR methinks.

src/main/java/fact/io/hdureader/BinTable.java
MaxNoe: It is very often used by people how are storing astronomical data like star catalogues (and fits actually defines those units).

I think at this point, we should not really bother. But if we want to make this a package on its own, we might want to handle more data.

src/main/java/fact/io/hdureader/FitsStream.java
```diff
@@ -0,0 +1,111 @@
...
+        if (ztable) {
+            reader = ZFitsHeapReader.forTable(eventsTable);
+        } else {
+            reader = BinTableReader.forBinTable(eventsTable);
+        }
+    }
+
+
+
+    private void applyDrsOffsetCalib(int numSlices, short[] data, short[] startCellData, short[] calibrationConstants){
```
dneise: One can save time here by not adding these constants here at all but instead adding them together with the other constants in the "other" DRS amplitude calibration which comes after reading the raw data. 
Not sure how much time it saves though. 
However, by not adding these constants, this "raw data" is already "pseudo calibrated" which means, people can already see single pulses on this pseudo calibrated data. Entirely for free.
By adding these constants, one just adds noise to the timeline, i.e. one scrambles the raw data ... which costs time and doesn't add value. 

One does "restore" the raw data, yes ... but who is interested in that noise raw data? 
I never met anybody.

MaxNoe: Oh, I never realised that. Wouldn't it make sense to calculate DRS Calibration constants on the the values in the fz file?

dneise: I don't understand.

MaxNoe: I mean, if the "Data" values in the fz files are allready roughly calibrated (how by the way?), doesn't it make sense to only calculate the residual calibration constants from these values?

This would spare us undoing a calibration to apply another one, right?

dneise: Yes, it would spare us one addition per pixel and cell. (at construction time, not at run time)

If we would however keep calling it "drs offset" or "drs baseline" it would be not comparable to what other people (e.g. those using MARS) or from other experiments call "drs offset", that I think is a major drawback.

 ... also storing the offsets, we call "BaselineMean" in units of the ADC would spare us one multiplication with a fixed number (at run-time), but we don't do that.

I think, this PR in particular is not the right place to discuss this and I also think the time to make things right is long gone. 

But I might be wrong .. I certainly don't want to stand in progress' way.

PR #180, opened 2016-12-20
Moving to streams-1.0.0

cbockermann:
Moved streams dependencies to de.sfb876.streams-0.9.26, updated @Parameter->@Service annotations where required.

All unit tests successful.


MaxNoe:
@mackaiver Should comment on this, because I think he tried  the newest stream version at some point but found a blocker.

mackaiver:
As far as I can see those issues have been fixed. I think this can be merged. Should be a major version release though I agree.

MaxNoe:
Today in the C3 Meeting we discussed that we wait for streams 1.0, which should be out tomorrow, and then update this PR to go to 1.0

cbockermann:
checking on it.

PR #181, opened 2016-12-20
Use new hdu reader for the aux service

MaxNoe:
Switch to new hdu reader in the aux service


dneise:
do you relly want to merge this into hdu_reader? not into the master? why not?

MaxNoe:
hdu_reader in master, so yes, I want to merge it into hdu_reader

Issue #183, opened 2017-01-16
single pulse extractor output truncation option


relleums:
The output of the single pulse extractor has artefacts in its beginning and its end.
Currently (fact-tools 0.17.2 spe pass2), and there is no air shower signal beyond slice ~ 130.
We want an option to truncate the output of the extractor, to remove artifacts and reduce output size. 
The truncation shall be controllable via parameters for the extraction processor. Defaults might be start: 15ns (30slices) and end: 65ns (130slices). 

All videos of the photon stream I have shown yet were reduced to 15ns-65ns, or even less anyhow.

| |   | start slice | end slice | remarks |
|---|---|---|---|---|
| exists | raw ROI | 0 | 300 | |
| exists | 1pe extraction range | 20 | 245 | truncate time marker and analog artifacts in the beginning |
| proposal | 1pe output range | 30 | 130 | truncate 1pe extraction artifacts and focus on air shower photons |

dneise:
solved by #185

Issue #182, opened 2017-01-16
single pulse extractor output is not yet size efficient, baselines too fine


relleums:
File size comparison for FACT run 20151001_011 with fact-tools 0.17.2 pass2    

| file type | size [MB] | reduction [rel.] |
| ------------- | ------------- |-----------|
| zfits  | 2935  | 1.0 |
| fact-tools jsonl.gz  | 305 | 9.6 |
| baseline in doubleSignDigits = 3 jsonl.gz  | 170 | 17.3 |
| optimized jsonl.gz  | 154 | 19.1 |
| optimized and truncated time lines jsonl.gz  | 94 | 31.2 |
| optimized, truncated time lines, and no baselines jsonl.gz  | 69 | 42.5 |

The baselines (useful for time series reconstruction) are a mayor contribution to the output size as they are stored as floats:
```"PhotonArrivalsBaseLine":[-0.9346934899690481,-0.914864300604518, ...```

This may be optimized by storing the baseline in e.g. centi 1pe amplitudes which is still fine enough:
```"PhotonArrivalsBaseLine100":[-93,-91, ...```

dneise:
started working on it, c.f https://github.com/fact-project/fact-tools/commit/da29965ddc243c01f6140dc4aa5754ca6e0ea935

MaxNoe:
The facttools JSONWriter has the option `doubleSignDigits` which sets the number of decimal places for doubles.

dneise:
yes, but this is globally affecting the precision, right? The proposal is to only change this for the BaseLines

MaxNoe:
Yes, i just wanted to mention it in case you were not aware of the option. (And I don't think we have anything that would require the default of 8 digits)

dneise:
closed by #186

PR #184, opened 2017-01-16
Singlepulseextractor output control

dneise:
solves #182 and #183

### DIFF
--- a/src/main/java/fact/extraction/SinglePulseExtraction.java
+++ b/src/main/java/fact/extraction/SinglePulseExtraction.java
@@ -95,7 +95,7 @@ public class SinglePulseExtraction implements Processor {
 
         double[] numberOfPulses = new double[npix];
         int[][] pixelArrivalSlices = new int[npix][];
-        int[] baseLine = new int[npix];
+        int[] baseLines = new int[npix];
         double[][] timeSeriesAfterExtraction = new double[npix][];
 
         SinglePulseExtractor.Config config = new SinglePulseExtractor.Config();
@@ -126,7 +126,7 @@ public class SinglePulseExtraction implements Processor {
             timeSeriesAfterExtraction[pix] = ElementWise.multiply(
                 result.timeSeriesAfterExtraction,
                 config.factSinglePeAmplitudeInMv);
-            baseLine[pix] = (int) Math.round(result.timeSeriesBaseLine() * 100);
+            baseLines[pix] = (int) Math.round(result.timeSeriesBaseLine() * 100);
         }
 
         addStartSliceOffset(pixelArrivalSlices);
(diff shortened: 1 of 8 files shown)


jebuss:
singlepeextactor

Here is a screenshot of the current effect off the signal reconstruction. Looks good, if you ask me!

src/main/java/fact/extraction/SinglePulseExtraction.java
```diff
@@ -13,6 +13,23 @@
...
+ * |                                                                       |
+ * |         |................ extraction window .................|        |
+ * |         | <---------- length = 225 ------------------------> |        |
+ * |         |                                                    |        |
+ * |         |       |.. output window ..|                        |        |
+ * 0        20       | <- length=100 ->  |                       245      300
+ *                  30                  130
+ *
  * Created by Sebastian Mueller
  * and some modifications from Jens Buss
```
jebuss: add dneise here

src/main/java/fact/features/singlePulse/timeSeriesExtraction/SinglePulseExtractor.java
```diff
@@ -46,10 +46,28 @@ public int numberOfPulses() {
...
         public double timeSeriesBaseLine() {
-            DescriptiveStatistics statistics = 
+            DescriptiveStatistics statistics =
                 new DescriptiveStatistics(timeSeriesAfterExtraction);
             return statistics.getMean();
-        }    
+        }
+
+        /**
+         * @param: startSlice is with respect to extraction window start.
```
MaxNoe: This javadocstring is not correct. See http://www.oracle.com/technetwork/articles/java/index-137868.html

src/main/java/fact/extraction/SinglePulseExtraction.java
```diff
@@ -21,8 +38,8 @@
     private String dataKey = null;
 
     @Parameter(
-        required = true, 
-        description =   "output key, a list of lists of arrival slices of "+ 
+        required = true,
+        description =   "output key, a list of lists of arrival slices of "+
```
jebuss: This not a list of lists anymore, its a 2dim in array

jebuss:
Well, at first I thought it is ok to use this multiplication for the sake of reducing the size of a single value instead of all values in the data item. But, with a second thought, I have to agree to what Max said. Its really not that intuitive and maybe we don't need to do this for a single variable, since we probably won't use the single pe extraction together with other feature extractions where we need a higher significance than two digits.

dneise:
Situation pre this PR:

 - keyname: "BaseLine"
 - unit: "Single Photon Amplitude"
 - number of keys with the same unit: 0

Situation after this PR:

 - keyname: "CentiBaseLines"
 - units: "1% of Single Photon Amplitude"
 - number of keys with the same unit: 0

MaxNoe:
Not the point. We always rage about contra-intuitive formats and units. Now we start the same here (without any documentation about the matter).

dneise:
I see your point.

> We could adapt the JSONWriter to allow it to set the significant digits per key not just globally. However, I think we can reduce the number of significant digits quite dramatically in general (4 digits should be enough for everything we write out).

This sounds like the way to go. I looked at the code ... doesn't look too easy ... in fact I would not even know where to start. 

At the moment I run a test with `doubleSignDigits = 3` ... just to see where we end up.

dneise:
I cherry picked the agreed upon parts into another branch and opened a dedicated PR for that.
see #185

dneise:
I added two processors which allow us to change the unit of Baseline outside of the processor creating Baseline in another PR. They are pretty general, so I guess it makes sense to give them their own PR.

see #186 

So this PR can be closed now.

PR #186, opened 2017-01-17
Add some numpy to xml

dneise:
Add two more processors which make the fact-tools XML feel a bit  more like numpy.

Now one can multiply a vector with a scalar (only if the vector is in fact a `double[]`)
and one can cast that `double[]` to an `int[]`.

I agree, it's not much and one would get this for free in a different environment, but its a step.

This PR is in connection to #182, since these two processors allow us to reduce the precision of the Baseline so it does not hurt that much anymore to write this noise to disk.

### DIFF
--- a/src/main/java/fact/utils/CastDoubleArrayToIntArray.java
+++ b/src/main/java/fact/utils/CastDoubleArrayToIntArray.java
@@ -23,8 +23,9 @@ public class CastDoubleArrayToIntArray implements Processor {
         double[] doubleArray = (double[]) input.get(inputKey);
         int[] intArray = new int[doubleArray.length];
 
-        for (int i=0; i<intArray.length; ++i)
+        for (int i=0; i<intArray.length; ++i){
             intArray[i] = (int) Math.round(doubleArray[i]);
+        }
 
         input.put(outputKey, intArray);
         return input;
--- a/src/main/java/fact/utils/CastDoubleArrayToIntArray.java
+++ b/src/main/java/fact/utils/CastDoubleArrayToIntArray.java
@@ -1,0 +1,40 @@
+package fact.utils;
+
+import stream.Data;
+import stream.Processor;
+import stream.annotations.Parameter;
+
+public class CastDoubleArrayToIntArray implements Processor {
+
+    @Parameter(
+        required = true,
+        description = "Key to your double array."
+    )
+    private String inputKey;
+
+    @Parameter(
+        required = true,
+        description = "Key to the output integer array."
+    )
+    protected String outputKey;
+
+    @Override
+    public Data process(Data input) {
+        double[] doubleArray = (double[]) input.get(inputKey);
+        int[] intArray = new int[doubleArray.length];
+
+        for (int i=0; i<intArray.length; ++i)
+            intArray[i] = (int) Math.round(doubleArray[i]);
+
+        input.put(outputKey, intArray);
+        return input;
+    }
+
+    public void setInputKey(String inputKey) {
+        this.inputKey = inputKey;
+    }
+    public void setOutputKey(String outputKey) {
+        this.outputKey = outputKey;
+    }
+
+}
--- a/src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
+++ b/src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
@@ -15,7 +15,7 @@ public class ElementwiseMultiplyDoubleArray implements Processor {
 
     @Parameter(
         required = true,
-        description = "Key to the output integer array."
+        description = "Key to the output double array."
     )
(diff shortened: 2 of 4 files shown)


src/main/java/fact/utils/CastDoubleArrayToIntArray.java
```diff
@@ -0,0 +1,40 @@
...
+        description = "Key to the output integer array."
+    )
+    protected String outputKey;
+
+    @Override
+    public Data process(Data input) {
+        double[] doubleArray = (double[]) input.get(inputKey);
+        int[] intArray = new int[doubleArray.length];
+
+        for (int i=0; i<intArray.length; ++i)
```
MaxNoe: Heartbleed anyone?

src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Parameter(
+        required = true,
+        description = "Key to your double array."
+    )
+    private String inputKey;
+
+    @Parameter(
+        required = true,
+        description = "Key to the output integer array."
```
MaxNoe: I think the output is still a double array.

src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
```diff
@@ -0,0 +1,46 @@
+package fact.utils;
+
+import stream.Data;
+import stream.Processor;
+import stream.annotations.Parameter;
+import fact.features.singlePulse.timeSeriesExtraction.ElementWise;
```
MaxNoe: I think it's clear now that this class belongs in a more general package than `singlePulse`

dneise: Wrong place for this comment? 
I guess you are taking up on a discussion you had with the author of Elementwise?

src/main/java/fact/utils/CastDoubleArrayToIntArray.java
dneise: no entiendo

src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
MaxNoe: Also: Does `Elementwise` create a copy? If not, the input array is `overwritten` in the sense that the operation happens inplace as the data item contains a reference to the array, which got multiplied by 5.

dneise: `Elementwise.multiply` does not act in place

src/main/java/fact/utils/CastDoubleArrayToIntArray.java
MaxNoe: Ah, sorry, wasn't heartbleed. It was `goto fail` https://blog.codecentric.de/en/2014/02/curly-braces/.

The missing braces thingy :) I just don't like `if`s and `for`s without braces

src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
MaxNoe: Maybe wrong place, maybe not. You started using a method outside its first intended scope ;) So it is this code point where it got evident for the first time.

src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
MaxNoe: Nice

src/main/java/fact/utils/CastDoubleArrayToIntArray.java
dneise: Ah ... the missing brace was not visible in the snippet above the comment... 
thx.

src/main/java/fact/utils/ElementwiseMultiplyDoubleArray.java
dneise: So what you telling me is, I should in this PR move `Elementwise` somewhere else so I can use it in its intended scope? Am I understanding you correctly? 

You said "this class" above .. and I was not sure which class you are talking about. 
Your comment strongly felt, like it was part of a discussion which I missed. 

So please make yourself clear here. Should I touch this other code within this PR? Is this necessary to make this PR get your approval, or not?

MaxNoe: I allready approved ;)

I had the discussion with @relleums that his ElementWise functions might be so usefull, that they should go in a more general place. And here is the first class, which uses them outside the singlePulse package.

So we should think about moving them.

PR #185, opened 2017-01-17
Single pulse extractor issue 183

dneise:
fix only #183

by cherry-picking the (non disputed) commits you have already seen in #184.

I added a changelog entry as well and bumped the version.

regarding #182 I don't know if bumping the version was premature. please comment.

PR #188, opened 2017-01-17
add two processors converting photon stream to arrival time and #photons

dneise:
We discussed yesterday, that the baseline might not be necessary at all, and we might be able to study this, if we converted the photon stream back to the two most basic observables:

 * arrival time per pixel and
 * number of photons per pixel

Here is a very basic proposal for those processors... I am currently looking into them ... i.e. trying them out in the viewer.

However ... early feedback is much appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 #Changelog for the fact-tools
 
+# Version 0.17.4 -- 17.01.2017
+
+* add fact.utils.PhotonStream2ArrivalTime
+* add fact.utils.PhotonStream2NumberOfPhotons
+
 # Version 0.17.3 -- 17.01.2017
 
 * output window: Cut out artifacts of SinglePulseExtractor.
--- a/src/main/java/fact/utils/PhotonStream2ArrivalTime.java
+++ b/src/main/java/fact/utils/PhotonStream2ArrivalTime.java
@@ -33,6 +33,9 @@ public class PhotonStream2ArrivalTime implements Processor {
         Arrays.sort(pulses);  // acts in place, but I guess this is no problem.
 
         int len = pulses.length;
+        if (len == 0){
+            return 0; // The plot does not like Double.NaN, so I use 0.
+        }
         if (len%2 == 1 ){
             median =  pulses[(len - 1) / 2];
         } else {
--- a/src/main/java/fact/utils/PhotonStream2ArrivalTime.java
+++ b/src/main/java/fact/utils/PhotonStream2ArrivalTime.java
@@ -23,11 +23,15 @@ public class PhotonStream2ArrivalTime implements Processor {
         double[] arrivalTimes = new double[singlePulses.length];
 
         for (int pix = 0; pix < singlePulses.length; pix++) {
-            DescriptiveStatistics stat = new DescriptiveStatistics();
-            for (int slice: singlePulses[pix]){
-                stat.addValue((double) slice);
+            if (singlePulses[pix].length == 0){
+                arrivalTimes[pix] = 0.; // use zero instead of NaN - plotter likes no NaN
+            } else {
+                DescriptiveStatistics stat = new DescriptiveStatistics();
+                for (int slice: singlePulses[pix]){
+                    stat.addValue((double) slice);
+                }
+                arrivalTimes[pix] = stat.getPercentile(50);
             }
-            arrivalTimes[pix] = stat.getPercentile(50);
         }
         input.put(arrivalTimeKey, arrivalTimes);
         return input;
--- a/src/main/java/fact/utils/PhotonStream2ArrivalTime.java
+++ b/src/main/java/fact/utils/PhotonStream2ArrivalTime.java
@@ -1,0 +1,52 @@
+package fact.utils;
+
+import java.util.Arrays;
+import stream.Data;
+import stream.Processor;
+import stream.annotations.Parameter;
+
+public class PhotonStream2ArrivalTime implements Processor {
+    @Parameter(
+        required = true,
+        description = "The arrival slices of the single pulses.")
+    private String singlePulsesKey = null;
+
+    @Parameter(
+        required = true,
+        description = "The reconstruted arrival time")
+    private String arrivalTimeKey = null;
+
+    @Override
+    public Data process(Data input) {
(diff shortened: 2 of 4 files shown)


dneise:
Played with it using this XML
```
<container>
  <properties url="classpath:/default/settings.properties" />
  <!-- Pathes to the input files -->
  <property name="infile" value="file:src/main/resources/testDataFile.fits.gz" />
  <property name="drsfile" value="file:src/main/resources/testDrsFile.drs.fits.gz" />
  <property name="integralGainFile" value="classpath:/default/gain_sorted_20131127.csv" />
  <property name="pixelDelayFile" value="classpath:/default/delays_lightpulser_20150217.csv" />
  <property name="auxFolder" value="file:src/main/resources/aux/" />

  <service id="auxService" class="fact.auxservice.AuxFileService" auxFolder="${auxFolder}" />
  <service id="calibService" class="fact.calibrationservice.ConstantCalibService" />

  <stream id="fact" class="fact.io.zfits.ZFitsStream"  url="${infile}"/>

  <!-- Description of the process and the corresponding stream -->
  <process id="2" input="fact">

    <fact.utils.PreviousEventInfo startCellKey="StartCellData" outputKey="prevEvents" limitEvents="10" />
    <include url="classpath:/default/data/calibration.xml" />
    <include url="classpath:/default/data/extraction.xml" />
    <include url="classpath:/default/data/cleaning.xml" />
    <include url="classpath:/default/data/pedestalParameters.xml" />

    <!-- Single Pe Extraction -->
    <fact.extraction.SinglePulseExtraction
            dataKey="DataCalibrated"
            outputKey="PhotonArrivals"
    />

    <fact.utils.PhotonStream2ArrivalTime
      singlePulsesKey="PhotonArrivals"
      arrivalTimeKey="doms_funny_arrival_time"
    />

    <fact.utils.PhotonStream2NumberOfPhotons
      singlePulsesKey="PhotonArrivals"
      arrivalTimeKey="doms_funny_arrival_time"
      numberOfPhotonsKey="doms_funny_num_phot"
    />

    <!-- Data Time Series Reconstruction from Single Pe Data -->
    <fact.utils.ConvertSinglePulses2Timeseries
            singlePulsesKey="PhotonArrivals"
            baseLineKey="PhotonArrivalsBaseLine"
            timeSeriesKey="DataCalibrated"
    />

    <!-- Extraction -->
    <!-- BasicExtraction, RisingEdgeForPositions, RisingEdgePolynomFit, TimeOverThreshold,
    PhotonChargeTimeOverThreshold, HandleSaturation, CorrectPixelDelays-->
    <include url="classpath:/default/data/extraction.xml" />
    <!-- Output: photoncharge, arrivalTime -->



    <fact.ShowViewer key="DataCalibrated" />
  </process>
</container>
```

dneise:
arrival times

2017-01-17-113457_1920x1080_scrot

dneise:
photon charges.
2017-01-17-113536_1920x1080_scrot

dneise:
The arrival times are not so back ... they just suffer from the artifacts at the edges of the extraction window, I guess ... for this see #185

MaxNoe:
Looks good!

For the times: Maybe a truncated mean with the ability to set  the borders?

src/main/java/fact/utils/PhotonStream2ArrivalTime.java
```diff
@@ -0,0 +1,55 @@
...
+        double[] arrivalTimes = new double[singlePulses.length];
+
+        for (int pix = 0; pix < singlePulses.length; pix++) {
+            arrivalTimes[pix] = this.getMedian(singlePulses[pix]);
+        }
+        input.put(arrivalTimeKey, arrivalTimes);
+        return input;
+    }
+
+    public double getMedian(int[] pulses){
```
jebuss: There is a very nice class for this, called DescriptiveStatistics. It can calculate all sort of statistical representations from a given Double array

dneise:
Maybe ... question is as always: 
 * Is that already optimization (Don't wanna do that now)? 
 * Or needed to show that the principle works?

Unclear to me ... I will test it against 0.17.3 once that is merged...

src/main/java/fact/utils/PhotonStream2ArrivalTime.java
dneise: Ah I see .. I just Ctrl-f searched the whole fact-tools code for the word "median" and found such code in `MuonRingClean.java` so I assumed this is the way to go in fact-tools.

dneise: very nice you say ... I see it can only work with `double[]`, so I have to convert that array again ... does not sound very nice to me.

jebuss: fact.util.Utils.toDoubleArray could do the job

MaxNoe:
```
for(Serializable value: item.get('inputKey')) {
     stats.addValue((double) value);
}
```

MaxNoe:
```
for(Serializable value: item.get('inputKey')) {
     stats.addValue((double) value);
}
```

MaxNoe:
```
for (int value: singlePETimeSlices[pixel]){
    stats.addValue((double) value);
}
```

dneise: thanks did it like max proposed.

dneise:
using 0.17.3

photon_charge

photon_charge_vs_doms_funny_num_photos

arrival_Time

arrivaltime_vs_doms_funny_arrival_time

dneise:
Here is a comparison similar to the one, made by jens ... just on much lower statistics
https://github.com/dneise/spe_study_comparision/blob/master/little_overview.ipynb

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 #Changelog for the fact-tools
 
+# Version 0.17.4 -- 17.01.2017
+
+* add fact.utils.PhotonStream2ArrivalTime
```
jebuss: Idont like the location fact.utils. Can we move this to its own package fact.photonStream?

Issue #195, opened 2017-01-17
Time over threshold pixel set is only written to the datatiem if empty


relleums:
```TimeOverThreshold``` writes the int array ```totPixelSet``` only to the data item if ```totPixelSet.length != 0```.
However, I think it does not harm to put it there all the time since empty arrays are very small (memory size).
The issue  #190 can be solved when we could always export this ```totPixelSet```. But therefore we need also to export it everytime, also when it is  ```totPixelSet.length == 0```

relleums:
Solved with commit 44459d4bbb78da5ebdba126ffa543121598bba39

Issue #190, opened 2017-01-17
Export list of saturated pixels in the single pulse extraction passes


relleums:
(Partly on behalf of Adrian)

We propose a variable length list in the output which stores the ids of saturated pixels.
Such a list would be most often empty (size efficient).

```"SaturatedPixels":[81,123,1439] ```

For the single pulse extractor passes we want to not run the pulse extraction on exceptional bright events e.g. 27s events or LIDAR events due to run time/wall time issues. Adrian proposes to not extract single pulses when a certain fraction of the pixels is saturated (maybe 10%?).

Adrian highly recommends to store such a list in the output of the single pulse extraction passes. Although these are again detector specific values, the list would at least be empty for most events. This way one can decide if one wants to process events with saturated pixels. Further we might prevent running into the wall time. Bright events were no extraction was performed simply do not have a photon_stream in them.

Could somebody point us to a sample run with  exceptional bright events?

relleums:
This turns out to be a duplicate or extension to issue #150

dneise:
This guy already outputs such a list...
but it is not used anymore at all... does this mean it does not work?

https://github.com/fact-project/fact-tools/blob/master/src/main/java/fact/extraction/HandleSaturation.java

see also #192

dneise:
Ah no ... this `TimeOverThreshold` guy seems to output such a list by the "timeOverThresholdSet" key.

dneise:
In the XML I would propose to do it like this:

    <fact.pixelsets.Length
      pixelSetKey="timeOverThresholdSetOverLay"
      outputKey="NumberOfSaturatedPixels"
    />
    <stream.flow.If 
        condition="%{data.NumberOfSaturatedPixels} &lt; 100"
    >
        <fact.extraction.SinglePulseExtraction
            dataKey="DataCalibrated"
            outputKey="PhotonArrivals"
        />
    </stream.flow.If>

MaxNoe:
This was solved in the commits linked above

PR #196, opened 2017-01-17
always export the totPixelSet, solve #195 and partly solves #190

relleums:
I do not think this will harm. What was the purpose of the if clause?


jebuss:
cant remember either

PR #201, opened 2017-01-19
AboveThresholdProcessor

relleums:
As we wished for. With basic unit tests.

### DIFF
--- a/src/main/java/fact/extraction/AboveThreshold.java
+++ b/src/main/java/fact/extraction/AboveThreshold.java
@@ -1,0 +1,72 @@
+package fact.extraction;
+
+import fact.Utils;
+import stream.Data;
+import stream.Processor;
+import stream.annotations.Parameter;
+import java.util.Arrays;
+import java.util.ArrayList;
+
+/**
+ * Finds pixels above a certain threshold limit. If only a single slice of the 
+ * pixel is above the threshold, this pixel is marked.
+ * Also the ratio of pixels above the threshold is calculated.
+ */
+public class AboveThreshold implements Processor {
+    @Parameter(
+        required = true, 
+        description = "A double array with length NROI times NPIX"
+    )
+    private String dataKey = null;
+
+    @Parameter(
+        required = true,
+        description = "A list of the pixels above the threshold. "+
+                      "Also the fraction of pixels above the threshold" 
+    )
+    private String outputKey = null;
+
+    @Parameter(
+        required = true,
+        description = "Threshold value to be exceeded"
+    )
+    protected short threshold = 1800;
+
+    @Override
+    public Data process(Data input) {
+
+        final int npix = (Integer) input.get("NPIX");
+        final int roi = (Integer) input.get("NROI");
+        final short[] timeSerieses = (short[]) input.get(dataKey);
+
+        ArrayList<Integer> pixelsAboveThresholdList = new ArrayList<>();
+        for (int pix=0; pix<npix; pix++) {
+            for(int slice=0; slice<roi; slice++) {
+                final int pos = pix*roi + slice;
+                if(timeSerieses[pos] > threshold) {
+                    pixelsAboveThresholdList.add(pix);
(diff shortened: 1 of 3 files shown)


src/main/java/fact/extraction/AboveThreshold.java
```diff
@@ -0,0 +1,73 @@
...
+    protected double threshold = 1800;
+
+    @Override
+    public Data process(Data input) {
+
+        final int npix = (Integer) input.get("NPIX");
+        final int roi = (Integer) input.get("NROI");
+        final short[] timeSerieses = (short[]) input.get(dataKey);
+        final short thresholdShort = (short)threshold;
+        ArrayList<Integer> pixelsAboveThresholdList = new ArrayList<>();
```
MaxNoe: You should use a pixelset here, not a plain int list

MaxNoe:
I changed the Processor to use a PixelSet and also save slice information, please have a look.

src/main/java/fact/extraction/AboveThreshold.java
```diff
@@ -37,25 +42,29 @@ public Data process(Data input) {
 
         final int npix = (Integer) input.get("NPIX");
         final int roi = (Integer) input.get("NROI");
-        final short[] timeSerieses = (short[]) input.get(dataKey);
+        final short[] timeSeries = (short[]) input.get(dataKey);
```
relleums: I can not wrap my head around this. There is no sense in a single time series for all the independent pixels. Why not having a plural 's'?

src/test/java/fact/extraction/AboveThresholdTest.java
```diff
@@ -39,10 +51,15 @@ public void testAboveThresholdCertainPixelsAbove() {
...
+        PixelSet testPixels = new PixelSet();
+        testPixels.addById(42);
+        testPixels.addById(314);
+        testPixels.addById(1337);
+
         for(int pix=0; pix<1440; pix++) {
             for(int slice=0; slice<300; slice++) {
                 final int pos = pix*300 + slice;
-                if(pix==42 || pix==314 || pix==1337) {
+                if((pix==42 || pix==314 || pix==1337) && (slice == 271 || slice == 256)) {
```
relleums: Looks good and simple to also test the slices.

relleums:
It looks like you managed to combine our demands. Since you told me that the PixelSet knows how to convert itself into a simple list when appended to the output json writer, I approve all this.

MaxNoe:
Series is both plural and singular. There is no word serieses


Am 23. Januar 2017 8:18:32 nachm. schrieb "Sebastian A. Mueller" 
<notifications@github.com>:

src/main/java/fact/extraction/AboveThreshold.java
MaxNoe: Series is both the singular and plural form. There is now serieses.

MaxNoe:
It's actually not the `PixelSet` itself but the `JSONWriter` who knows how to handle a `PixelSet` through a `TypeAdapter`;)

src/main/java/fact/extraction/AboveThreshold.java
relleums: I propose to replace all timeseries(es) with timelines  or sth. else which has a sane plural representation. We should not limit ourselves by using variable names which have no flexibility or restricted features.

src/main/java/fact/extraction/AboveThreshold.java
```diff
@@ -0,0 +1,82 @@
...
+                      "The fraction of slices above the threshold is put into outputKey + SliceRatio" +
+                      "The number of slices above the threshold is put into outputKey + SliceCount"
+    )
+    private String outputKey = null;
+
+    @Parameter(
+        required = true,
+        description = "Threshold value to be exceeded"
+    )
+    protected double threshold = 1800;
```
jebuss: set this to null

PR #202, opened 2017-01-20
Create photonstream package

### LINKED ISSUE #198: Create PhotonStream package
create a `fact.photonstream` package and move all processors that are related to the photon stream to this package, e.g.:

* `fact.extraction.SinglePulseExtraction`
* `fact.utils.PhotonStream2ArrivalTime`
* `fact.utils.PhotonStream2NumberOfPhotons`
* `fact.utils.ConvertSinglePulses2Timeseries`

dneise:
fixes #198


dneise:
should I bump the version and write something to the changelog?

Are you happy with the processors, which have been moved and their final position?

MaxNoe:
We have not yet released 0.18. I would put this in 0.18 as well.

relleums:
I would not merge it like this. Here ALL tools related with single pulses are moved into the ```fact.photonstream``` package. But some of those (mainly by Jens and Katie) are solving different tasks e.g. gain extraction. I propose these should stay in place in the existing ```fact.extraction.singlePulse```. We only move the ```fact.extraction.singlePulse.timeSeriesExtraction``` into the novel ```fact.photonstream```.

dneise:
I think somebody else, should take over, who knows, where should go what.

Issue #197, opened 2017-01-19
use "integralGains" for single pulse extraction?


dneise:
When extracting the `photoncharge` we use these individual pixel gain number:
https://github.com/fact-project/fact-tools/blob/master/src/main/resources/default/gain_sorted_20131127.csv
(btw. this is no CSV at all)
They look like this:
![integral_gains_from_file](https://cloud.githubusercontent.com/assets/8200858/22109521/63ba52b6-de58-11e6-822d-62b7f80c8a61.png)

I was wondering if we should also use these numbers in the single pulse extractor, here:
https://github.com/fact-project/fact-tools/blob/master/src/main/java/fact/extraction/SinglePulseExtraction.java#L117

At the moment the constant `factSinglePeAmplitudeInMv` which is used here, is 10.
but we can also convert these integralGains from the file into this number.
The distribution for all pixels looks like this then:
![conversion_factors](https://cloud.githubusercontent.com/assets/8200858/22109410/e836dcae-de57-11e6-8661-4cd0b738c706.png)

jebuss:
Good idea, makes sense to me. Let's test it. However the amplitude might only be a 10% effect at max.

dneise:
Is it a good idea then? Or a waste of time? We have very very limit amout of time only. 
The single photon extraction pass should have been started since long. 

Question is not whether its nice or not ... surely one can do it and test it ... but is it necessary? 
Will people say: "This extraction is invalid", if we do not use these individual gains?

jebuss:
I don't know if it is necessary or not. But I think it would not make the result worse, only better! That's my guess. So put it in, if you agree. If we don't use it, we might misestimate Cherenkov pulses by up to 10%, since we apply a 10mv Amplitude to a 9mv pulse, which is also my guess.

dneise:
> since we apply a 10mv Amplitude to a 9mv pulse, which is also my guess.

Have a look at the lower image: 
https://cloud.githubusercontent.com/assets/8200858/22109410/e836dcae-de57-11e6-8661-4cd0b738c706.png

It shows the factors, which would be applied instead of the constant factor 10. ...

jebuss:
Yes that's what I meant.

jebuss:
I was talking about the worst case which is the outlier at 9mv

jebuss:
To shorten the discussion. I think it's a good idea. I think it's worth it. I thinking would improve the result. I think we should go for it.

dneise:
[...]

PR #204, opened 2017-01-20
Individual pixel gain in spe

dneise:
fixing #197

### DIFF
--- a/src/main/java/fact/extraction/SinglePulseExtraction.java
+++ b/src/main/java/fact/extraction/SinglePulseExtraction.java
@@ -8,6 +8,7 @@ import stream.annotations.Parameter;
 import java.util.Arrays;
 import fact.features.singlePulse.timeSeriesExtraction.SinglePulseExtractor;
 import fact.features.singlePulse.timeSeriesExtraction.ElementWise;
+import fact.features.singlePulse.timeSeriesExtraction.TemplatePulse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


src/main/java/fact/extraction/SinglePulseExtraction.java
```diff
@@ -148,6 +170,42 @@ private void addStartSliceOffset(int[][] arr) {
         }
     }
 
+    public double[] loadIntegralGainFile(SourceURL inputUrl, Logger log) {
+        factSinglePePulseIntegral = TemplatePulse.factSinglePePulseIntegral();
```
dneise: this is shit .. I am misusing loadIntegralGainFile as an `__init__` method. so I initialize this member here...
but that's of course total bullshit.

dneise:
We should not merge this, if we think #207 is good, which I do.

MaxNoe:
Yep, let's wait for #207

MaxNoe:
There are conflicts and a lot of changes after the last review.

@jebuss @relleums How should we proceed?

MaxNoe:
@dneise @jebuss @relleums This has some overlap with #333 

* How should we proceed here?
* What is the difference between "integral gain" and "single pulse integral gain"

dneise:
> What is the difference between "integral gain" and "single pulse integral gain"?

I've got no idea

MaxNoe:
What we call Integral Gain in FACT-Tools is around 240, what is here called Single Pulse Integral gain seems to be around 1.0, so how come from one to the other?

PR #207, opened 2017-01-21
A Single pulse gain calibration service to solve issue #197

relleums:
I am afraid that the only (almost clean) way to have the gain for more than one processor is the introduction of a Single pulse gain calibration service. This service can now provide several processors. This is needed now, because now there are two processors needing the gain: ```BasicExtraction``` and ```SinglePulseExtraction```


src/main/java/fact/features/singlePulse/timeSeriesExtraction/TemplatePulse.java
```diff
@@ -65,4 +65,70 @@
...
+
+        final double samplingPeriodInNs = 0.5;
+        return factSinglePePulse(lengthInSlices, samplingPeriodInNs);
+
+    }
+
+    public static double factSinglePePulseIntegral() {
+
+        final double half_height = 0.5; // this is fix forever :-D
+        final double samplingPeriodInNs = 1e-3;
```
MaxNoe: We have 2 GSample/s, right?

src/main/java/fact/calibrationservice/SinglePulseGainCalibService.java
```diff
@@ -0,0 +1,69 @@
...
+import stream.annotations.Parameter;
+import stream.io.SourceURL;
+import stream.io.CsvStream;
+import stream.Data;
+import fact.Constants;
+
+/**
+ *
+ **/
+public class SinglePulseGainCalibService implements CalibrationService {
```
MaxNoe: indentatio is off in this file

src/main/java/fact/extraction/BasicExtraction.java
```diff
@@ -31,9 +30,8 @@
 	protected String outputKeyMaxAmplPos = null;
 	@Parameter(required = true, description="outputKey for the calculated photoncharge")
 	protected String outputKeyPhotonCharge = null;
-	
-	@Parameter(required = false, description = "The url to the inputfiles for the gain calibration constants",defaultValue="file:src/main/resources/defaultIntegralGains.csv")
-	protected SourceURL url = null;
+    @Parameter(required = true, description = "The calibration service for the integral single pulse gain")
```
MaxNoe: This should be `@Service`, see #180

src/main/java/fact/extraction/SinglePulseExtraction.java
```diff
@@ -80,14 +85,39 @@
     )
     protected int extractionWindowLengthInSlices = 225;
 
+    @Parameter(
```
MaxNoe: This should be `@Service`, see #180

dneise:
Okay, I was working on this a bit, but during integration of the processor only I realized that calibrating the amplitudes to units of 'SinglePhotonAmplitudes' (instead of mV) does not play well with something like the saturation processor. 

Maybe however it makes sense to let the ADC saturation processor operate in the realm of voltages and only after this processor calibrate the time lines to 'SinglePhotonAmplitudes'.

The fix pixel delays should be treated much earlier I think. Within the resampling of the timelines sounds reasonable to me.

Until we didn't discuss this, I will pause working on this PR.

PR #228, opened 2017-05-03
Add method to get all table names in hdureader

MaxNoe:
Returns a `String[]` with all table names.


mackaiver:
I'd rather make the `hdunames` thing public. Maybe rename it to something more useful.
What do you think?

MaxNoe:
I'm open to that. I'm not that fluent on "How to Java API"

Issue #222, opened 2017-02-01
Update dependency of org.reflections library.


mackaiver:
The newest version is  [0.9.10](https://search.maven.org#artifactdetails|org.reflections|reflections|0.9.10|jar) as far as I know. We are on 0.9.9 in the master branch.

MaxNoe:
done in #233

Issue #221, opened 2017-02-01
Update Guava


jebuss:
Update the guava dependency to the newest version, if possible. Check what has to be adapted

````

    <dependency>
      <groupId>com.google.guava</groupId>
      <artifactId>guava</artifactId>
      <version>17.0</version>
    </dependency>
```

MaxNoe:
Done in #233

Issue #220, opened 2017-02-01
Get rid of old libraries, that are already in java 8


jebuss:
In our pom.xml we have dependencies that are already included in java8 in some way, e.g. 

```
    <dependency>
      <groupId>joda-time</groupId>
      <artifactId>joda-time</artifactId>
      <version>2.7</version>
    </dependency>
````
got through the pom.xml, find out wich of them are already in Java8, delete them one by one and fix the upcoming problems.

MaxNoe:
done in #233

PR #229, opened 2017-05-09
Refactor

s58237:
Issue #220 were implemented.

Issue #221 and #222 were realized and other libaries were updated.

Settings.properties and Settings_mc.properties file were merged in one file ("settings.properties) and the keys were summerized.

examples/stdAnalysis/data/analysis.xml and examples/stdAnalysis/mc/analysis_mc.xml were merged in one file "analysis.xml". The characteristic "fTotalEnergy" in the data deciedes wheter a data or an mc-analysis is executed.When fTotatlEnergy==null then a data-analysis is choosen.

Code of xml-files in src/main/resources/default is inserted directly in files of example an the xml files werde deleted.


src/main/java/fact/auxservice/AuxPoint.java
```diff
@@ -12,11 +13,11 @@
...
  */
 public class AuxPoint implements Comparable<AuxPoint>{
-    private DateTime timeStamp;
+    private OffsetDateTime timeStamp;
     private ImmutableMap<String, Serializable> data;
 
-    public AuxPoint(DateTime timeStamp){
-        this.timeStamp = new DateTime(timeStamp);
+    public AuxPoint(OffsetDateTime timeStamp){
+        this.timeStamp =OffsetDateTime.of(timeStamp.toLocalDate(),timeStamp.toLocalTime(), ZoneOffset.of("+00:00"));
```
mackaiver: Why is it neccessary to create a new object here? It's basically a copy isn't it?

src/main/java/fact/auxservice/AuxPoint.java
```diff
@@ -25,17 +26,17 @@ public AuxPoint(DateTime timeStamp){
      * @param timeStamp the timestamp specifying  when the auxiliary data was recorded.
      * @param data a map containing the names and values from the .fits files as key values pairs.
      */
-    public AuxPoint(DateTime timeStamp, Map<String, Serializable> data){
+    public AuxPoint(OffsetDateTime timeStamp, Map<String, Serializable> data){
         this.data = new ImmutableMap.Builder<String, Serializable>().putAll(data).build();
-        this.timeStamp = new DateTime(timeStamp);
+        this.timeStamp =OffsetDateTime.of(timeStamp.toLocalDate(),timeStamp.toLocalTime(), ZoneOffset.of("+00:00"));
```
mackaiver: see above

MaxNoe:
I think this is too large to be reviewable and too much can break.

Can we have only the library concerned changes first? These are important to get fast into master and should be easy to review. 

Also today we spoke about the question `OffsetDateTime` vs `ZonedDateTime` and concluded that we should use `ZonedDateTime` everywhere which is just minor changes.

MaxNoe:
I opened a PR with only the library changes

MaxNoe:
I don't like that these containers with the standard calibration and extraction stuff got removed. I like these building blocks for ease of use.

MaxNoe:
I'm closing this for now as there is too much to be effectively reviewed and to much conflicts to the master.

PR #233, opened 2017-05-24
Library updates

MaxNoe:
I took the library versions and JODA Time updates related commits from here #229 
and put them in their own branch to make it easier to review.

### DIFF
--- a/src/main/java/fact/auxservice/SqliteService.java
+++ b/src/main/java/fact/auxservice/SqliteService.java
@@ -2,9 +2,6 @@ package fact.auxservice;
 
 import com.google.common.cache.*;
 import fact.auxservice.strategies.AuxPointStrategy;
-import org.joda.time.*;
-import org.joda.time.format.DateTimeFormat;
-import org.joda.time.format.DateTimeFormatter;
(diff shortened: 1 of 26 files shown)


src/main/java/fact/auxservice/SqliteService.java
```diff
@@ -119,15 +114,31 @@ public int hashCode() {
...
+
+                        String earlierTime=time.minusHours(4).toString();
+                        String [] dateAndTime=earlierTime.split("T");
+                        String [] timeSplitZone=dateAndTime[1].split("Z");
+                        String earlier=dateAndTime[0]+" "+timeSplitZone[0].substring(0,8);
+
+                        String laterTime=time.plusMinutes(window).toString();
+                        dateAndTime=laterTime.split("T");
+                        timeSplitZone=dateAndTime[1].split("Z");
+                        String later = dateAndTime[0]+" "+timeSplitZone[0].substring(0,8);
```
mackaiver: I don't understand why we cannot use a format string here?
Not very important since this class is obsolete in my opinion anyways.

PR #237, opened 2017-07-04
Reading in Corsika Run Number and/or particle ID from Run header information

relleums:
To read the CORSIKA run number, the 2nd fits output file of CERES with the CORSIKA run header must be read. The reading is done using the CeresStream. I think this is needed in any case for all simulation processing with fact-tools. Otherwise the fact-tools output can neither contain the CORSIKA RUN number nor the primary particle ID. When not relying on external bookkeeping (e.g. in the photon-stream), the Run number is mandatory and must be propagated.


jebuss:
Hi,
Looks generally good to me. However, I have two questions. Concerning the resource files. Do we need to have the root files in the resources? The information in them is somewhat redundant and we agreed that we never want to read root files into fact-tools. 

Concerning the reader: What do we do in case the _RunHeader file is missing. I guess in this case the run is failed anyways, since important information is lost. However, the fact that two files are need is certainly hidden in the code and thus maybe a surprise for the user. Not sure how to solve this. Either a specific property in the XML (which is possibly redundant) or a meaningful warning if the _RunHeader file is missing.

src/main/java/fact/io/CeresStream.java
```diff
@@ -0,0 +1,66 @@
...
+            expandedURL = new URL(this.url.getProtocol(), this.url.getHost(), this.url.getPort(), this.url.getFile());
+        }
+
+        fitsstream = new FITSStream(new SourceURL(expandedURL));
+        fitsstream.init();
+
+        Path path = Paths.get(expandedURL.getPath());
+        String runHeaderFileName = path.getFileName().toString().replace("_Events", "_RunHeaders");
+
+        Path runHeaderPath = Paths.get(path.getParent().toString(), runHeaderFileName);
```
jebuss: I suggest to check here if the _RunHeaders file exists and in case it does not, to raise an exception

src/test/java/fact/io/CeresStreamTest.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Created by Kai on 27.06.17.
+ */
+public class CeresStreamTest {
+
+    @Test
+    public void testFitsStream() {
+
+        try {
+            URL u =  FITSStreamTest.class.getResource("/ceres_output/18000/00018000.000_D_MonteCarlo018_Events.fits.gz");
```
jebuss: can we may be use the mc file from the resources or replace it with this one and an according runHeaders file. I think we shouldn't blow up the jar with too many fits files in the resources

src/test/java/fact/io/CeresStreamTest.java
```diff
@@ -0,0 +1,62 @@
...
+            e.printStackTrace();
+            fail("Could not read FitsFile");
+        }
+    }
+
+    @Test
+    public void testRunHeaderKeys() {
+
+        try {
+            URL u =  FITSStreamTest.class.getResource("/ceres_output/18000/00018000.000_D_MonteCarlo018_Events.fits.gz");
```
jebuss: same here

src/test/java/fact/io/CeresStreamTest.java
mackaiver: yeah. maybe. as long as it stays in test/resources it wont get packaged into the jar anyways.

mackaiver:
> Do we need to have the root files in the resources?

No. It was just copied to have the same file structure as produced by ceres. I can replace them by dummy files.

relleums:
A consequence of the user specifying the external ceres output directory instead of the ```Events```file in this directory would be that the interface for simulation processing changes and tools like e.g. ERNA would have to adopt accordingly. Yet another possibility would be to let the user explicitly provide both the ```Events``` and the ```RunHeader``` file, but this needs even more adjustments in the tools like ERNA, and is in my opinion bad and unnecessary as the ceres output directory structure is very predictable. What do you think? 

- Add an exception message in case the ```RunHeader``` file does not exist when the user specifies the ```Events``` file?

or

- Let the user specify the surrounding ceres output directory instead and not throw custom exceptions in case expected files do not exist?

MaxNoe:
Any news here?

src/main/java/fact/io/CeresStream.java
MaxNoe: Won't this be done by the next line?

Issue #236, opened 2017-06-04
Delta uses atan instead of atan2


MaxNoe:
Not sure if this is an issue.

MaxNoe:
I checked, as the ellipse is symmetrical, there is no reason to use atan2.

MaxNoe:
Thinking about directional reconstruction, I came to the conclusion, that this is important because only with the full angular information one can make full use of the m3 moment information. Because only with the direction of the eigenvector, the meaning of m3's sign is clear.

MaxNoe:
Thinking again, this is not the case. Closing for now

Issue #232, opened 2017-05-23
Alpha uses atan instead of atan2


MaxNoe:
THis might lead to wrong results

MaxNoe:
I came to the conclusion, that alpha needs to use atan2, as `Distance` and `Alpha` are basically polar coordinates of the source with respect to the cog.

MaxNoe:
As we probanly won't calculate source dependent parameters in FACT-Tools anymore I'll close this issue

PR #242, opened 2017-08-08
Use atan2 for alpha and delta

MaxNoe:
See #236 and #232

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <groupId>de.sfb876</groupId>
   <artifactId>fact-tools</artifactId>
   <name>fact-tools</name>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -4,7 +4,7 @@
   <groupId>de.sfb876</groupId>
   <artifactId>fact-tools</artifactId>
   <name>fact-tools</name>
-  <version>0.18.1</version>
+  <version>0.18.1-SNAPSHOT</version>
```
jebuss: shall it be a snapshot?

MaxNoe:
Discussed with Key about this and checked the return values of the Eigenvector calculation. This is not necessary

PR #246, opened 2017-10-13
Noise generator

tarrox:
A lot of changes preparing for the samplePedastalEvent function. Most of it is independent of this and just extends functionality, most often making it more adjustable.

* DrsCalibration now is able to reverse its doing
* DrsCalibration added a new test for the reverse feature
* DrsCalibration now has the option to set the key with the drsfilename in it
* DrsTimeCalibration now allows to set the keys to be used
* InterpolateTimeSeries now allows to set the UnixTimeUTC key
* Fixed an Error in SourcePosition where it returned Integer instead of doubles in one of the possible paths
* HDUReader Added a skipToRow function to the Reader and implemented it for both bintable an zfits
* HDUReader Added test for skipToRow
* TimeSeriesFeatures Added the ability to actually substract the avarageData
* CombineDataArray Added a new Prozessor to combine (add,sub,mul,div) 2 double arrays.
* Remaping Added the ability to reverse the remapping of the data and added a testcase for it


src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -128,39 +136,54 @@ public Data process(Data data) {
...
+			output = new double[rawfloatData.length];
 		}
 
 		double[] calibrated = applyDrsCalibration(rawfloatData, output,
 				startCell);
-		data.put(outputKey, calibrated);
+		if (!reverse)
+			data.put(outputKey, calibrated);
+		else {
+			short[] shortres = new short[calibrated.length];
```
jebuss: Can you use something more explainatory than shortres

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -256,11 +279,21 @@ public Data process(Data data) {
 				triggerOffsetPos = pixel * drsTriggerOffsetMean.length / 1440
 						+ slice;
 
-				vraw = data[pos] * dconv;
-				vraw -= drsBaselineMean[offsetPos];
-				vraw -= drsTriggerOffsetMean[triggerOffsetPos];
-				vraw /= drsGainMean[offsetPos];
-				vraw *= 1907.35;
+				if (!reverse) {
```
jebuss: shouldn't `reverse` be handed over to the function. I know its a global variable. But i think, what it means for the function, is more obvious if its handed over

src/main/java/fact/datacorrection/DrsTimeCalibration.java
```diff
@@ -41,11 +41,13 @@
     SourceURL url = new SourceURL(DrsTimeCalibration.class.getResource("/long_term_constants_median.time.drs.fits"));
 
 
-    @Parameter(required = false, description = "")
+    @Parameter(required = false, description = "The name of the key, where the finished time calibrated data items are inserted.")
     String outputKey = "DataCalibrated";
 
+    @Parameter(required = false, description = "The name of the key with the startcelldata")
```
jebuss: Out of curiosity: why do you need to define the startcell key?

src/main/java/fact/datacorrection/InterpolateTimeSeries.java
```diff
@@ -35,6 +35,12 @@
     private String dataOutputKey = null;
     @Parameter(required = false, description = "The minimum number of neighboring pixels required for interpolation", defaultValue="3")
     private int minPixelToInterpolate = 3;
+
+	@Parameter(required = false, description = "The key for the resulting badPixelSet.")
```
jebuss: Why do you need the badPixelSet key? It's not used.

src/main/java/fact/io/hdureader/BinTableReader.java
```diff
@@ -70,6 +72,14 @@ public static BinTableReader forBinTable(BinTable binTable){
             return  map;
         }
 
+        @Override
+        public void skipToRow(int num) throws IOException {
```
jebuss: docu?

src/main/java/fact/io/hdureader/FITSStream.java
```diff
@@ -110,4 +113,17 @@ public Data readNext() throws Exception {
 
         return item;
     }
+
+    public Data readNextRaw() throws Exception {
```
jebuss: docu

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -118,6 +122,29 @@ public static ZFITSHeapReader forTable(BinTable binTable){
         return getNextRow(false);
     }
 
+    @Override
+    public void skipToRow(int num) throws IOException {
+        //calc zshrink problem
+        //num = num+1;
+        int numSkip = num-(num%zshrink); // the amount we can skip in the catalog due to zshrink
+        int skipMore = num - numSkip; // num%zshrink works too, the amount we have to skip with getNext
```
jebuss: rather call it `skipLength` or `numRowsToSkip` ?

src/main/java/fact/statistics/TimeseriesFeatures.java
```diff
@@ -32,8 +32,9 @@
 	private int searchWindowRight = 250;
 	@Parameter(required=true,description="name of the key of the calculated features")
 	private String outputKey = null;
-	
-	
+	@Parameter(required=false, description="Whether to do the calculation with substraction of the avarageData.")
```
jebuss: Maybe explain, what this will mean for the features. 
Use longer name: e.g. `subtractSmoothedData`

src/main/java/fact/utils/CombineDataArrays.java
```diff
@@ -0,0 +1,101 @@
...
+    @Parameter(required = true, description = "The key to your first array.")
+    private String firstArrayKey;
+    @Parameter(required = true, description = "The key to your second array.")
+    private String secondArrayKey;
+    @Parameter(required = false, description = "The key for the resulting array.")
+    private String outputKey;
+    @Parameter(required = true, description = "The operation to perform, (add, sub, mul, div)")
+	private String op;
+    
+	private int npix = 1440*300;
```
jebuss: can these magic numbers be avoided

src/main/java/fact/utils/CombineDataArrays.java
```diff
@@ -0,0 +1,101 @@
...
+		Utils.isKeyValid(input, firstArrayKey, double[].class);
+		Utils.isKeyValid(input, secondArrayKey, double[].class);
+
+		double[] resultArray =  new double[npix];
+
+		double[] array1 	 = (double[]) input.get(firstArrayKey);
+		double[] array2 	 = (double[]) input.get(secondArrayKey);
+
+		if (op.equals("add")) {
+			for (int pix = 0; pix < npix; pix++) {
```
jebuss: instead of using npix you could make this processor generic looping over array elements, but then you need to have a sanity check if both arrays have the same length. But this is allready implemented in `fact.photonstream.timeSeriesExtraction.ElementWise`

src/main/java/fact/datacorrection/DrsTimeCalibration.java
tarrox: Because i need to calibrate the pedestal and it has the key ped_startcelldata

src/main/java/fact/utils/CombineDataArrays.java
MaxNoe: it's not even the number of pixels

src/main/java/fact/io/hdureader/BinTableReader.java
tarrox: In the reader interface, should I copy it also here?

src/main/java/fact/datacorrection/InterpolateTimeSeries.java
jebuss: solved

tarrox:
Added some fixes for some files.
Added the SamplePedestalEvent processor and a xml file for doing the sampling.
Now everything I needed is here.

PR #247, opened 2017-10-17
renames db to aux_source

jebuss:
since with the current version of the aux services also a directory tree is supported


jebuss:
this adapts for the changes in https://github.com/fact-project/erna/pull/43 and shouldn't be merged before the PR in erna is merged

MaxNoe:
But one has to change the class to fit the input right? Here the sqlite service is used in this example.

PR #258, opened 2017-11-15
Write timestamp in std analysis

### LINKED ISSUE #250: Write out timestampe
We should convert our `UnixTimeUTC` tuple to a timestamp and write it out as ISO String

MaxNoe:
Fixes #250


mackaiver:
can we do something about MC timestamps here? Or maybe somewhere else.

Issue #74, opened 2015-07-27
Possible bug in ConcentrationCore.java


mackaiver:
Originally reported by: **Dominik Neise (Bitbucket: [dneise](http://bitbucket.org/dneise), GitHub: [dneise](http://github.com/dneise))**

---

It seems, there was a bug in the MARS implementation of calculating, what we call the shower core concentration. The bug was apparently spotted and solved over the last weekend.

I guess, the same 'bug' should also occur in ConcentrationCore.java. 

The MARS fix was done here:
https://trac.fact-project.org/changeset/18275
if you want to look at it.

Maybe it's not wise to jump at the fix and implement it as well, but rather take the chance to deal also with the comment on line 65 of ConcentrationCore.java.
Also one must keep in mind, that the trunk version of MARS, is not the "official" version, which is used to publish results, but rather the development version. So it might well be, that this is just a test of sorts.

---
- Bitbucket: https://bitbucket.org/cbockermann/fact-tools/issue/74

MaxNoe:
I think we should just reimplement ConcCore how we deem it correctly

MaxNoe:
Fixed in #261

PR #261, opened 2017-11-15
Reimplement conc core

MaxNoe:
This reimplements ConcCore without copying the MARS code verbatim and without sticking to 1 character variable names

### DIFF
--- a/src/main/java/fact/Utils.java
+++ b/src/main/java/fact/Utils.java
@@ -349,22 +349,17 @@ public class Utils {
 	 * @param delta
 	 * @return an array having two elements {l, t}
 	 */
-	public static double[] transformToEllipseCoordinates(double x, double y,
-			double cogX, double cogY, double delta) {
+	public static double[] transformToEllipseCoordinates(double x, double y, double cogX, double cogY, double delta) {
 		double translatedX = x - cogX;
 		double translatedY = y - cogY;
 
-		double dist = Math.sqrt(translatedX * translatedX + translatedY
-				* translatedY);
+		double sinDelta = Math.sin(delta);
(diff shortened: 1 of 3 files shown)


MaxNoe:
No. See issue #74 . We do not really expect to have the same results. 

I implemented these feauture how it is described in the comments.

MaxNoe:
Comparison plots new and old on a Crab data run with ~18000 Events:

comparison1d
comparison2d
I think the results are similar enough and the Code is much easier to understand now.

dneise:
Could you point to these conments in Mars you are refering to? Or, even better: point to the place in fact tools, where you explain what you define as ConcCore?

MaxNoe:
> Could you point to these conments in Mars you are refering to? Or, even better: point to the place in fact tools, where you explain what you define as ConcCore?

In Fact-Tools this is now in the javadocs of the processor:
https://github.com/fact-project/fact-tools/pull/261/files#diff-d2718c8d2e104928d3e18a591f16e42aR38

In Mars it's here:

https://trac.fact-project.org/browser/trunk/Mars/mimage/MHNewImagePar.h#L31

dneise:
Ah okay, I get it now .. I must say I mixed the word "ConcCore" up with "ConcCOG" since Core is also a word which expresses something "centered" ... reading then that you count all pixels with a Maha-dist < 1 did not help me, since I did not know what you used for the covariance matrix ... 

"inside the Hillas ellipsis" should have told me what was going on ... 

I know you are not responsible for the term "ConcCore" which was basically what puzzled me and reading the docstring did not help me, but sent me to wikipedia, to study the Maha-dist, just to learn, that I am missing the information about the cov-matrix.

PR #263, opened 2017-11-16
Add trigger emulation

jebuss:
This is an implementation of an Emulation of the Trigger used in FACT. I intend to use this to search for data and MC mismatches on Trigger level.

This package supports:

* signal shaping
  * Shaping only on the valid range of the time series (exclude noisy ranges)
  * shape pixel by pixel (don't shift full time series)
* summation of pixel's time series for each patch
* discrimination of the patch signals
* a pseudo ratescan on the patches
* handles stars in FOV and excludes their pixels them from the trigger decission
* handles faulty pixels and excludes them from the trigger decission
* N-out-of-4 logic
* N-out-of-40 logic

Feel free to comment!


jebuss:
This is an example of a ratescan on the example data file in the fact_tools resources
pseudo_ratescan

Blue is the sum of all patches, orange is just pixel 0

src/main/java/fact/Constants.java
```diff
@@ -10,4 +10,6 @@
 
 	public static final double PIXEL_SIZE = 9.5; //9.5mm
 
+
+	public static final double MILLIVOLTPERDAC = 610.*Math.pow(10,-3);
```
MaxNoe: Add underscores between words here `UPPERCASE_WITH_UNDERSCORES`

MaxNoe: Also `610e-3`

dneise:
what is the x-axis in the plot above?

jebuss:
> what is the x-axis in the plot above?
Damn, yes, forgot that. Added it to the post.

dneise:
Thanks, how did you get the threshold in DAC values?

jebuss:
Found this in Voglers Thesis: Vogler, P. E. (2015). Design and commissioning of the trigger electronics for a novel Geiger-mode avalanche photodiode based camera for Imaging Atmospheric Cherenkov Telescopes. https://doi.org/10.3929/ETHZ-A-010568419

jebuss:
Uhhh, wait, im stupid. Plotted wrongly. X-axis is the bin number. I did not identify it with the DAC value of the bin

dneise:
Erm ... what do you mean by bin ? (maybe you want to update that again in the picture and I just delete all my questions here, so the chat is cleaned up again.

jebuss:
> Erm ... what do you mean by bin ? (maybe you want to update that again in the picture and I just delete all my questions here, so the chat is cleaned up again.

Done, we can clean up

jebuss:
@dneise @MaxNoe: I implemented everything that I wanted to implement for this package. If tests pass it should be ready for review.

mackaiver:
Hey. It builds on my machine. But it takes more than 10 minutes. Maybe that's why travis is failing?

src/main/java/fact/TriggerEmulation/Ratescan.java
```diff
@@ -0,0 +1,217 @@
...
+
+            ratescanResult.n_primitives_arr.add(n_tiggered_patches);
+            ratescanResult.triggerPrimitives_arr.add(currentTriggerPrimitives);
+            ratescanResult.patchTriggerSlices_arr.add(patchTriggerSlice);
+
+        }
+        return ratescanResult;
+    }
+
+    public static long lin2log(int z) {
```
MaxNoe: docstring, what is this for?

src/main/java/fact/TriggerEmulation/Ratescan.java
```diff
@@ -0,0 +1,217 @@
...
+                    skipLast
+            );
+
+            n_tiggered_patches = countPrimitives(currentTriggerPrimitives);
+
+            log.debug("Threshold: "+threshold+" nPrimitives: "+n_tiggered_patches);
+
+            ratescanResult.n_primitives_arr.add(n_tiggered_patches);
+            ratescanResult.triggerPrimitives_arr.add(currentTriggerPrimitives);
+            ratescanResult.patchTriggerSlices_arr.add(patchTriggerSlice);
```
MaxNoe: Combination of snake and camel case

src/main/java/fact/TriggerEmulation/Ratescan.java
```diff
@@ -0,0 +1,217 @@
...
+
+        public int[][] getTriggerSlices(){
+            int[][] triggerSlices = new int[1][1];
+            return patchTriggerSlices_arr.toArray(triggerSlices);
+        }
+    }
+
+    public int countPrimitives(boolean[] primitives){
+        int counts = 0;
+        for (boolean primitive:
```
MaxNoe: weird line break

src/main/java/fact/TriggerEmulation/SumUpPatches.java
```diff
@@ -0,0 +1,129 @@
...
+    private PixelSet invalidPixels = new PixelSet();
+
+    @Override
+    public Data process(Data item) {
+        Utils.isKeyValid(item, key, double[].class);
+        double[] data = (double[]) item.get(key);
+
+        int roi = (int) item.get("NROI");
+
+        int n_patches = Constants.N_PIXELS/Constants.N_PIXELS_PER_PATCH;
```
MaxNoe: Maybe add a constant for this?

src/main/java/fact/TriggerEmulation/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * @author Jens Buß, jens.buss@tu-dortmund.de
+ * This package collects processors to emulate the hardware trigger and trigger logic of FACT.
```
MaxNoe: Maybe add more info about the trigger here?

jebuss:
Ready for review @dneise, @MaxNoe, @mackaiver

src/main/java/fact/features/source/PixelSetForSourcePosition.java
```diff
@@ -0,0 +1,99 @@
...
+
+            CameraPixel starPixel = pixelMap.getPixelBelowCoordinatesInMM(starPosition.xMM, starPosition.yMM);
+
+            if (starPixel == null) {
+                log.debug("Star not in camera window");
+                continue;
+            }
+            starSet.add(starPixel);
+
+            for (CameraPixel px : pixelMap.getNeighborsForPixel(starPixel)) {
```
MaxNoe: Unnecessary complicated, a `CameraPixel` knows its position, so why give the id to the function and do the lookup again?

Also `CameraCoordinate` has a `euclideanDistance` method, so you could do `starPosition.euclideanDistance(new CameraCoordinate(px.getXPosititionInMM(), px.getYPositionInMM())` 

Even better would maybe be to directly add a `position` member to `CameraPixel`

src/main/java/fact/Constants.java
```diff
@@ -1,7 +1,13 @@
 package fact;
 
 public class Constants {
+    public static final double MILLIVOLT_PER_DAC = 610e-3;
```
dneise: where does this number come from again? could you cite something here?

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,116 @@
+package fact.TriggerEmulation;
+
+import fact.Constants;
+
+public class Discriminator {
```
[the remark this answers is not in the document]
dneise: by discriminator you mean something like a Schmitt-Trigger? 
https://de.wikipedia.org/wiki/Schmitt-Trigger

When looking up discriminator one finds something like this:
https://de.wikipedia.org/wiki/Diskriminator

I think we learned the word discriminator in or detector lecture back in the days ... right? But I wonder if the word is easily understood by many people

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,120 @@
...
+     *Discriminate the signal of a given patch
+     * @param data timeseries, array[n_pixels_pe_patch]
+     * @param threshold threshold of the discriminator in DAC
+     * @param minTimeOverThreshold minimum time the signal has to stay above the threhold
+     * @param skipFirst number of slices to ignore at the beginning of the time series
+     * @param skipLast number of slices to ignore at the end of the time series
+     */
+    public static int discriminatePatch(
+            double[] data,
+            int threshold,
```
dneise: you convert `threshold` from dac to mV further down, so this `threshold` could be named `threshold_in_dac` to make the unit clear for an outside user of this function.

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,120 @@
...
+            int minTimeOverThreshold,
```
dneise: just as above, you might want to give a hint here about the time unit you are having in mind?

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,120 @@
...
+            int skipFirst,
+            int skipLast
+    ) {
+        int counter = 0;
```
dneise: maybe rename this to explain *what* is being counted here?

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ *Class with functions to discriminate the signals of a given patch (or a bunch of them).
+ * Furhtermore it contains helper functions for the discrimination e.g. conversion between DAC and mV
+ */
+public class Discriminator {
+
+    public static int default_slice = 0;
+
+    /**
+     *Discriminate the signal of a given patch
```
dneise: maybe explain better what the return value of this function is ... maybe like:
"""
Return the slice (index of data) of the first occurence of:
  * signal is over threshold *and*
  * signal is over threshold for a long enough time.
"""
or so.

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param data timeseries, array[n_pixels_pe_patch]
```
dneise: I'm sorry .. didn't read java for some time .. by `array[n_pixels_pe_patch]` you mean: 

This is an array and the length is equal to the number of pixels we have in a patch? Which is 9, right?
In the implementation this looks like a it is a timeseries with a length of maybe 300 or so.... 
confused.

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,120 @@
...
+            if (counter >= minTimeOverThreshold){
+                return patchTriggerSlice;
+            }
+        }
+        return default_slice;
+    }
+
+    /**
+     * loop over pixels patch by patch and discriminate each patch
+     * @param data array[n_patches][n_pixels_pe_patch]
```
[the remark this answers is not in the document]
dneise: I assume a typo?

src/main/java/fact/TriggerEmulation/Discriminator.java
dneise: this counter counts the time the signal was over threshold in units of slices. So maybe: `time_over_threshold`?

dneise:
Found another reason for sticking to a certain line length limit:

If you want your colleagues to review your code ... give them a chance to actually read it

screenshot from 2018-10-10 09-34-54

src/main/java/fact/TriggerEmulation/EmulateDiscriminator.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+    public int default_slice = Integer.MAX_VALUE;
+
+    @Override
+    public Data process(Data item) {
+        double[][] data = (double[][]) item.get(key);
+
+        int n_patches = Constants.N_PATCHES;
+        double millivoltPerDAC = Constants.MILLIVOLT_PER_DAC;
```
dneise: I think this variable is defined but never used, some kind of static code analysis tool might be better at spotting such cases than a human tho.

src/main/java/fact/TriggerEmulation/EmulateDiscriminator.java
```diff
@@ -0,0 +1,120 @@
...
+        int[] patchTriggerSlice = new int[n_patches];
+
+        boolean[] triggerPrimitives = discriminatePatches(
+                data,
+                n_patches,
+                patchTriggerSlice,
+                threshold,
+                minTimeOverThreshold,
+                skipFirst,
+                skipLast
```
dneise: I believe the 4-tuple consisting of `threshold, minTimeOverThreshold, skipFirst, skipLast` forms some kind of a group in large portions of this code, so maybe one could tidy up by creating a little class/struct for these 4?

dneise:
I think this is all I have for the moment. .. just comments on form not on content.

I think reviewing the test cases would be interesting. I mean there is a gazillion ways to implement a trigger which is "somehow similar to FACT, but does of course not include all the nasty real life features" ... the devil is in the details. And maybe it would be cool to somehow summarize which parts of the real features have been taken into account and which parts have been omitted since they were either not important enough (using physicists intuition to decide what is important and what not, of course) or were simply too time consuming to be taken into account. 

All is fine, but if someone in the future thinks: Hey maybe we see this feature, since the trigger was not simulated detailed enough. That it would certainly be nice to give them a point to see what's missing and what not.

src/main/java/fact/TriggerEmulation/Discriminator.java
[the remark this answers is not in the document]
jebuss: yes its a typo. should be `per`

jebuss:
Hey Guys, thanks for the review. Sorry, coming back to it with delay. I try to go through it and answer your questions and adapt to your requests

src/main/java/fact/TriggerEmulation/Discriminator.java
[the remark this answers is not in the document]
jebuss: Thats a good question. You are right, I used the term we used back in the physics lab courses and I meant a device like this: http://web.physics.ucsb.edu/~phys128/experiments/muonphysics/Instrument%20manuals/LeCroy%20623B%20data%20sheet.pdf. However we are also using this term in FACTs design paper, which is why I used it here. 
Not sure, what would be a more meaningful term.

src/main/java/fact/TriggerEmulation/Discriminator.java
jebuss: nice catch, that was a copy&paste failure. Yes, you are right. It's the lenght is number of time slices of the series. Since this more confusing than helping I delete this.

src/main/java/fact/TriggerEmulation/EmulateDiscriminator.java
jebuss: Yes, was needed before. Not needed anymore

jebuss:
> And maybe it would be cool to somehow summarize which parts of the real features have been taken into account and which parts have been omitted since they were either not important enough (using physicists intuition to decide what is important and what not, of course) or were simply too time consuming to be taken into account.
> All is fine, but if someone in the future thinks: Hey maybe we see this feature, since the trigger was not simulated detailed enough. That it would certainly be nice to give them a point to see what's missing and what not.

That's a fair point. I'll try to do that. I will have to wrap up the trigger emulation for my thesis anyway. Once I am done with that I will put it into the code.

jebuss:
uuups, accidentally closed the PR

src/main/java/fact/TriggerEmulation/EmulateDiscriminator.java
jebuss: Not sure if this makes sense. In the end it is just four variables each being an int. However, `skipFirst` and `skipLast` are used in many different processors in order to define a valid operation window on the time series. Maybe we should open an issue to add a class `ValidTimeSeriesWindow` and refactor all processors using a valid window. I think it would even make sense to serialize this class and hand it to the data item so other processors working with a window can used hat specific window too. What do you think.

jebuss:
@dneise @MaxNoe I tried to fix/apply you requests apart from the summary about what is not yet implemented and the class holding the discriminator input parameters. Do you think this is ready to merge, can we live without the two mentioned things or is some crucial missing?

examples/triggerEmulation/find_max_possible_threshold.xml
```diff
@@ -0,0 +1,75 @@
...
+    <include url="classpath:/analysis/cleaning.xml" />
+
+    <fact.filter.ShapeSignal
+            key="DataCalibrated"
+            outputKey="DataShaped"
+            shift="20"
+            factor="0.96"
+    />
+
+    <fact.features.source.PixelSetForSourcePosition
```
MaxNoe: I think this should use the new `StarService` and probably be called `PixelSetForStarPosition`

examples/triggerEmulation/ratescan.xml
```diff
@@ -0,0 +1,79 @@
...
+    <include url="classpath:/analysis/pedestalParameters.xml" />
+
+    <fact.filter.ShapeSignal
+            key="DataCalibrated"
+            outputKey="DataShaped"
+            shift="20"
+            factor="0.96"
+    />
+
+    <fact.features.source.PixelSetForSourcePosition
```
MaxNoe: See above

src/main/java/fact/TriggerEmulation/Discriminator.java
```diff
@@ -0,0 +1,160 @@
...
+            int thresholdInDAC,
+            int minTimeOverThreshold,
+            int skipFirst,
+            int skipLast)
+    {
+
+        DiscriminatorOutput[] results = new DiscriminatorOutput[data.length];
+
+        for (int patch = 0; patch < data.length; patch++) {
+            results[patch] =
```
MaxNoe: strange linebreak here

src/main/java/fact/TriggerEmulation/EmulateLogic.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * sorted insert of int value into an int array
+     * @param x
+     * @param list
+     */
+    public static void insertToArrayListSorted(int x, ArrayList<Integer> list) {
```
MaxNoe: This silently ignores the case `>0`, maybe it should throw an error?

src/main/java/fact/TriggerEmulation/EmulateLogic.java
```diff
@@ -0,0 +1,159 @@
...
+        int pos = Collections.binarySearch(list, x);
+        if (pos < 0) {
+            list.add(-pos-1, x);
+        }
+    }
+
+    /**
+     * Check if nOutOf40 triggerPrimitive where in the given time window
+     * @param list
```
MaxNoe: What is list?

src/main/java/fact/TriggerEmulation/EmulateLogic.java
```diff
@@ -0,0 +1,159 @@
...
+     * @return
+     */
+    public static boolean areEnoughTriggersInOneTimeWindow(ArrayList<Integer> list,
+                                                           int nOutOf40, int timeWindowSize){
+        for (int i = 0; i <= list.size() - nOutOf40; i++) {
+            /**TODO: This does not distinguish between patches from the same FTU, so it may happen that nOutOf40 or more
```
MaxNoe: There is still a todo here. If this function is just valid for 1 out of 40, this should raise an error for other cases.

src/main/java/fact/TriggerEmulation/EmulateLogic.java
jebuss: ah I see

src/main/java/fact/features/source/PixelSetForSourcePosition.java
```diff
@@ -0,0 +1,83 @@
...
+import stream.Data;
+import stream.Processor;
+import stream.annotations.Parameter;
+
+
+/**
+ * generate a pixel set from the coordinates of stars in the camera given by the starPositionKeys
+ * Created by jbuss on 30.08.18.
+ */
+public class PixelSetForSourcePosition implements Processor {
```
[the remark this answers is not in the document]
jebuss: could you guide me to the place where you use the `StarService` because in the cleaning the `StarPositionsKey` is still used:
https://github.com/fact-project/fact-tools/blob/7c675d6ea21ec9b2f1f0e854984b571a3c02a1dc/src/main/java/fact/cleaning/TwoLevelTimeNeighbor.java#L60-L61

jebuss:
Allright. Please have a look. I hope I did everything you guys requested. I also added a small summary about what I know that is missing that is in the real Trigger but not in the sofware trigger. You can find it in the package info

jebuss:
> Well .. it would be crucial, if we wanted to answer this question: "Does this PR emulate the trigger within these requirements: [list, of, requirements]"

I added a small summary to the package info describing what is implemented and what is known to be missing. For sure I cannot say if I forgot about something there. But thats it as far as I can reconstruct it.

jebuss:
@dneise @MaxNoe ready to merge?

src/main/java/fact/TriggerEmulation/package-info.java
```diff
@@ -0,0 +1,71 @@
...
+ * <li>{@link fact.TriggerEmulation.Ratescan} is a processor that performs a software ratescan on summed patch time series
+ * by iterating over a list of thresholds and applying {@link fact.TriggerEmulation.Discriminator} to the data
+ * <li>{@link fact.TriggerEmulation.findMaximumTriggerThreshold}  is a processor that Scan the TriggerThreshold for each
+ *  event in order to find the maximum possible threshold to keep the event by applying
+ *  {@link fact.TriggerEmulation.Discriminator}.
+ * <li>{@link fact.TriggerEmulation.SumUpPatches} lets you sum up the signals of each patch element wise and
+ * returns an array of patchwise timeseries;
+ * </ul>
+ *
+ *
```
dneise: Could you elaborate a bit on the temporal behaviour here? The hardware trigger (at least in the so called "n-out-of-4" part) is a bit slower (or has a lower bandwidth) than the rest of the trigger electronics. This results in the trigger only firing if the signal is over threshold for a certain (though unknown) time. 

To what extent this behavious is simulated here, I do not know. 

Also I believe this behaviour can be emulated by shaping the signal with a lower bandwidth, so it might be emulated in `fact.filter.ShapeSignal`... but I do not know this.

src/main/java/fact/Constants.java
```diff
@@ -1,8 +1,20 @@
...
+    public static final int N_PIXELS_PER_PATCH =9;
     public static final double FOCAL_LENGTH_MM = 4889.0;
     public static final double PIXEL_SIZE_MM = 9.5;
+    public static final int N_PATCHES = N_PIXELS/N_PIXELS_PER_PATCH;
     public static final double FOV_DEG = 4.5;
+    /** Source: Vogler, Patrick E., DOI:10.3929/ETHZ-A-010568419,
+     *  URL: https://www.research-collection.ethz.ch/handle/20.500.11850/108302,
+     *  2015, P.56
+     */
+    public static final double MILLIVOLT_PER_DAC = 610e-3; //source 1 (below)
```
dneise: Source is above

src/main/java/fact/TriggerEmulation/package-info.java
dneise: Ah btw. this is only from my memory .. I do not know if Patrick has written something about this in his thesis

src/main/java/fact/Constants.java
jebuss:
```suggestion
    public static final double MILLIVOLT_PER_DAC = 610e-3; //source 1 (above)
```

src/main/java/fact/TriggerEmulation/package-info.java
jebuss: some if it he did. A time our threshold is considered. I added info about that and also a reference to his thesis

dneise: cool

Issue #252, opened 2017-11-09
HDU Reader probelm: fact-tools crashes when executing noise sampling


jebuss:
fact-tools crashes when executing noise sampling from PR #246. 

Seems this is a problem with the HDU reader.

Nov 09 2017 14:56:45 [INFO ] fact.datacorrection.DrsCalibration: Using .drs File /fact/raw/2013/12/04/20131204_007.drs.fits.gz
Nov 09 2017 14:56:58 [INFO ] fact.datacorrection.DrsCalibration: Using .drs File /fact/raw/2014/01/27/20140127_008.drs.fits.gz
Nov 09 2017 14:57:10 [ERROR] stream.runtime.AbstractProcess: Error occurred in process '1': java.lang.IllegalStateException: Duplicate key Key: (
^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@, Value: , Comment:
Nov 09 2017 14:57:10 [ERROR] stream.runtime.ProcessThread: java.lang.RuntimeException: java.lang.IllegalStateException: Duplicate key Key: (^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@, Value: , Comment:
        at fact.utils.SamplePedestalEvent.getNoiseEvent(SamplePedestalEvent.java:206)
        at fact.utils.SamplePedestalEvent.process(SamplePedestalEvent.java:82)
        at stream.runtime.AbstractProcess.process(AbstractProcess.java:126)
        at stream.runtime.AbstractProcess.execute(AbstractProcess.java:184)
        at stream.runtime.ProcessThread.run(ProcessThread.java:129)
Caused by: java.lang.IllegalStateException: Duplicate key Key: (^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@^@(^@^@
[...]

PR #264, opened 2017-11-21
Fixhdureader

tarrox:
Critical Fix to allow the reading of all of our fact data files. Basically implementing the multiple rows per TILE functionality as there is data that uses it. The data that fails is all before the 2013/10/19 (inclusive). Which is the data after which fact started using zfits instead of gz (all data on fhgfs is still in gz, isdc is fz, which is why we didn't catch it earlier). When they compressed them from gz into fz they choose some other zfits parameters (At least I think that happened, because i think the multiple row per tile is not really practical when taking data but can be used when recompressing without issues).

To achieve all this, this is what I did: 

* Implemented a caching mechanism to cache whole tiles not just singular rows.
* Changed the getNextRow method to use the cache instead of direct load
* Huffman decoding needed a little changing as each tile can have multiple huffmans
* Added some new exceptions to make finding missing features easier.

I also fixed some errors with the reader:

* The 2880 padding calculation was wrong, which was fixed
* The ZDrsCellOffsetTable is not always compressed -> Changed to general getNextRow


MaxNoe:
>  The data that fails is all before the 2013/10/19 (inclusive). 

This is not true. fact-tools `0.17.2` successfully analysed most of the data in 2013 on the isdc. 
Output of erna for all runs before 2013-12-31:
```
success              21675
failed                  27
walltime_exceeded       10
```

So what exactly does this fix?

jebuss:
> This is not true. fact-tools 0.17.2 successfully analysed most of the data in 2013 on the isdc.
Output of erna for all runs before 2013-12-31:

Which fits reader was used in this version and in erna?
The std_analysis.xml in version 0.17.2 still contains the old zFits reader and not the HDU reader, if I am not mistaken

MaxNoe:
Yes, I just realised that as well. Sorry. The `fact.io.zfits.ZFitsReader` was used.

jebuss:
fixes #252

MaxNoe:
But as far as I know, @relleums  used the hdureader to process the PhotonStream data. And he did encounter some problems which were already fixed. I know thought everything was working fine.

@relleums can you comment?

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -448,7 +518,9 @@ static BlockHeader fromBuffer(ByteBuffer buffer) throws IOException{
                     compression = Compression.RAW;
                     break;
             }
-            return new BlockHeader(size, order, compression);
+            // size(8 Bytes)+order(1 Bytes)+numberOfProcessings(1 Bytes)+Prozessings(2 Bytes*numberOfProcessings)
```
MaxNoe: typo in `Prozessings`

relleums:
Indeed, since about 1/2 year now I use the novel HduReader for the automatic photon-stream production at ISDC. fact-tools version is 0.18.1, commit 8a96abd8dc9e54ebfaedff97b093be72255d5b1f.

```xml
<stream id="fact" class="fact.io.hdureader.FITSStream" url="${infile}"/>
```

Most of the zfits files can be read. To be honest, I did not check if some files are still failing or what the exception messages are. I will look this up now.

jebuss:
Hey @relleums, thanks for the overview. I indeed forgot that you worked on that. If I understand you correctly, there might still be some edge cases were the HDU reader does not work, right?

So maybe this is what @tarrox found and hopefully fixed with this PR. 

What happened on @tarrox's side was, that he implemented a procedure in the HDU reader that allows to jump to a certain event in the zfits file (in our case interleaved pedestal events). In the first iteration, this failed for some files, because this method jumped to a wrong position in the fits file. This test was on a sample from Crab 2013/14. From this we concluded that there might be some edge cases were some zfits files behave differently to others. 

> I suggest to go on as I did before when deploying the photon-stream production. Apply the reader to be evaluated to a random 1% of our zfits runs scattered across all epochs from 2011 to 2018. Compare the number of processed (read in) events with the number of expected events (to be found in the run info table). 

Basically we did something alike by chance as we select certain interleaved pedestal events randomly that we tried to process. A difference is that we did not walk successively through the file. 

So, I think what is needed right now is an explantation of these edge cases and the according errors of the fact-tools process.

@tarrox, I suggest that you give some more details on what happened there, what went wrong and how you fixed it.

jebuss:
So in short: we need issues containing the error message

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -244,57 +288,76 @@ private void unsmooth(short[] data){
      * @param buffer the coplet tile buffer (with blockheaders, but without the tile headers.)
      * @throws IOException in case the buffer ends before all bytes are read
      */
-    private short[] huffmanDecompression(ByteBuffer buffer) throws IOException {
+    private short[] huffmanDecompression(ByteBuffer buffer, int numRows, int sizeSingleRow) throws IOException {
```
mackaiver: Not quite sure what this all does now. However I get the feeling this does the same for many rows instead of one. I do think its nicer when the loop happens outside of this function. Any reason why that won't work?

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -244,57 +288,76 @@ private void unsmooth(short[] data){
...
+                        }
                     }
-                }
 
-                ByteWiseHuffmanTree node = currentNode.children[q.peekByte()];
-                if (node.isLeaf) {
-                    symbols[index] = node.payload.symbol;
-                    index++;
-                    q.remove(node.payload.codeLengthInBits - depth*8);
+                    ByteWiseHuffmanTree node = currentNode.children[q.peekByte()];
```
mackaiver: In the diff it looks like the idendation is wrong.

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -402,11 +465,17 @@ static TileHeader fromStream(DataInputStream stream, boolean ignoreWrongHeader)
...
 
-        private BlockHeader(long size, String order, Compression compression) {
+        private BlockHeader(long size, String order, Compression compression, long sizeHeader) {
             this.size = size;
             this.order = order;
             this.compression = compression;
+            this.headerSize = sizeHeader;
+        }
+
+        public long getDataSize() {
```
mackaiver: missing javadocs

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
tarrox: It will work it is just that due to the way stuff was implemented before, this was the result, basically I would need to split the Huffman up a tiny little bit at the beginning. Will change this.

tarrox:
Baisically what happend is that i tried to read a file that our hdu reader couldn't read, the problem was with the reader and not the skipper (although it also needed a small update). The reason our reader can't read some of the files is due to the fact that the zfits-format allows for a variety of options how to save the data (for example combining several rows into a single tile), but our reader was written only implementing the options that are currently in use to save the files.

So to fix it I implemented some of the features that are used in some of the files, in this case the multi row per tile thing. I also added a few checks for some of the features that we don't support in case they were used so we get some nicer error messages.

The reason why it is not all implemented is because it creates a lot of complexity for something that isn't used anyways (And shouldn't be used, because some of this options don't give any benefit).

Also just for reminder our data on phido doesn't have this problems due to the fact that this data is still in gz format and not fz.

src/main/java/fact/io/hdureader/HDU.java
```diff
@@ -113,11 +113,7 @@ public BinTable getBinTable(){
      * @return the number of bytes to skip to hte next hdu.
      */
     long offsetToNextHDU(){
-        if(sizeOfDataArea() == 0){
-            return 0;
-        }
-        long numberOfBlocks = sizeOfDataArea() / 2880;
-        return 2880 * numberOfBlocks + 2880;
+        return ((sizeOfDataArea()+2879)/2880)*2880;
```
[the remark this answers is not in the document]
tarrox: This is written in the comment for the function. As this is just a one liner this should be fine, I think

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
tarrox: fixed

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
tarrox: fixed, now there is a new blockHuffmanDecompression that calls the huffmandecompression function for each row.

src/main/java/fact/ZFitsTester.java
```diff
@@ -0,0 +1,105 @@
...
+import java.net.MalformedURLException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ZFitsTester {
+    private static class FileInfo {
```
MaxNoe: Use GSON to write json

tarrox:
I fixed the given suggestions and added a new Class 'fact.ZFitsTester' this class can be executed on its own and it expects files as its parameters which are then send trough the hdu reader to test. At the end it creates a nice json file containing the filename, status, and info, where info is either the eventcount of the file or the error Message if the file couldn't be read for whatever reason.

To execute you just do:
'java -cp target/fact-tools-0.18.1-SNAPSHOT.jar fact.ZFitsTester pathToFz1 pathToFz2'

src/main/java/fact/ZFitsTester.java
tarrox: changed it

tarrox:
i now also created a testSubsetZFits.sh script you execute it with

```./testSubsetZFits.sh /fact/raw 100```

It will take a random of 100 zfits files from /fact/raw/*/*/*/*.fits.fz and process them trough the ZFitsTester.

**Comment by mackaiver** [comment]
> The reason why it is not all implemented is because it creates a lot of complexity for something that isn't used anyways (And shouldn't be used, because some of this options don't give any benefit).

so this only happens on certain pedestal files? Is that why it didn't appear before?

File: src/main/java/fact/io/hdureader/ZFITSHeapReader.java [hunk 7]
```diff
@@ -130,17 +132,58 @@ public static ZFITSHeapReader forTable(BinTable binTable){
      *
      * The old ZFits implementation by M.Bulinski ignored the problem by default.
      *
-     * @param ignoreWrongTileHeader if true, ognores wrong tileheader information.
+     * @param ignoreWrongTileHeader if true, ignores wrong tileheader information.
      * @return a map containing the the data from the rows columns.
      * @throws NoSuchElementException iff hasNext() is false
      */
     public OptionalTypesMap<String, Serializable> getNextRow(boolean ignoreWrongTileHeader) throws IOException {
```
mackaiver: So the problem is that there can be now many rows in one tile right? Why do we need to read many tiles then? Or is the problem that one row is split over many tiles?

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -130,17 +132,58 @@ public static ZFITSHeapReader forTable(BinTable binTable){
...
+            Buffer curBuffer = tileCache.get(column.name);
+
+            if (compressedBlocks.contains(column.name)) {
+                // Buffer is ShortBuffer
+                ShortBuffer shortBuffer = (ShortBuffer) curBuffer;
+                short[] data = new short[column.repeatCount];
+                shortBuffer.get(data);
+                map.put(column.name, data);
+            } else {
+                //Buffer is ByteBuffer
```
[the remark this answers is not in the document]
tarrox: The comment is more for the if else construct to know the state of the buffer, it is of course a little bit overkill due to the fact that i cast the buffer to the correct type directly afterwards. I will remove it and just make sure that the shortbuffer above is also the first line of code, so the code makes sense.

MaxNoe:
@tarrox @jebuss I think you still need to clarify why this is needed. As @relleums points out, the hdureader could read > 99 % of the files on the isdc without these changes.

Please explain why these changes are needed and which files are affected. 
The statement 
> The data that fails is all before the 2013/10/19

Is clearly wrong.

tarrox:
It is not wrong try e.g. file:
*/fact/raw/2013/10/01/20131001_011.fits.fz`
`/fact/raw/2013/04/10/20130410_019.fits.fz`

or any other files before the given deadline (the files on the fhgfs are still gz so no problem, but the isdc is different as they converted the files to fz), they all have this problem because they have `ZTILELEN=100`

you can check the headers with 
```
from astropy.io import fits
hdu = fits.open(path)
hdu[2].header['ZTILELEN']
```
if it says 100 it can't be read without this fix

jebuss:
> so this only happens on certain pedestal files? Is that why it didn't appear before?

No it happens for interleaved pedestal events in data files

jebuss:
> @tarrox @jebuss I think you still need to clarify why this is needed. As @relleums points out, the hdureader could read > 99 % of the files on the isdc without these changes.

> Please explain why these changes are needed and which files are affected.

So it appeared while jumping to certain (interleaved pedestal) events in a zfits file. As shown in #252 the reader arrived at a position where it was not able to interpret the content and failed. Maybe this is only an issue for interleaved pedestals or event jumping.

> When I tried to read files with the current master, it also failed for some runs i tried prior to 2013/10/19, but that does not fit @relleums observations.

Where did you get that > 99% of files could be read? 

> Most of the zfits files can be read. To be honest, I did not check if some files are still failing or what the exception messages are. I will look this up now.

This sounds to me that there is a chance that also 0.18 failed for some runs, but we did not notice it. Or it is due to the event jumping.

MaxNoe:
Ok I just talked with Sebastian on the phone. He was not sure that he used the hdureader for the files prior to 2013 something. But he has analysed a lot of files (even down to 2012) on the isdc. So we are still a bit confused. But I would go forward an merge this, as this is clearly an improvement over the master.

MaxNoe:
Can we add test files (Maybe truncated) for ZTILELEN > 1 und ZSHRINK > 1?

src/main/java/fact/io/hdureader/HDU.java
```diff
@@ -113,11 +118,7 @@ public BinTable getBinTable(){
      * @return the number of bytes to skip to hte next hdu.
      */
     long offsetToNextHDU(){
-        if(sizeOfDataArea() == 0){
-            return 0;
-        }
-        long numberOfBlocks = sizeOfDataArea() / 2880;
-        return 2880 * numberOfBlocks + 2880;
+        return ((sizeOfDataArea()+(FITS_BLOCK_SIZE-1))/FITS_BLOCK_SIZE)*FITS_BLOCK_SIZE;
```
jebuss: isn't this just the same as: sizeOfDataArea()+FITS_BLOCK_SIZE-1)
Or are there numerical reasons?

tarrox: numerical, all are ints

jebuss:
> Can we add test files (Maybe truncated) for ZTILELEN > 1 und ZSHRINK > 1?

+1

src/main/java/fact/io/hdureader/HDU.java
MaxNoe: Yes, you are correct. The `/ FITS_BLOCK_SIZE)*FITS_BLOCK_SIZE` should not be necessary and just add noise

tarrox:
> Can we add test files (Maybe truncated) for ZTILELEN > 1 und ZSHRINK > 1?

Working on it, do you have a ZSHRINK filename at your disposal or a way to find one?

src/main/java/fact/io/hdureader/HDU.java
MaxNoe: > numerical, all are ints

Can you explain? Why do you need to add not needed multiplications / divisions because they are ints?

MaxNoe: It's totally not obvious from the code. Add least split it to multiple lines and add comments to make clear what this is supposed to do.

MaxNoe:
> Working on it, do you have a ZSHRINK filename at your disposal or a way to find one?

Should be all files with > 30000 Events, so you could check the runinfo db

tarrox:
I just created both files, but didn't push them yet.
```
7.5M testDataZshrink.fits.gz  6 Events ZSHRINK=2
33M testDataZTILELEN.fits.fz 100 Events ZTILELEN=100 one tile
```
Is this ok? Especially the ZTILELEN is pretty big due to all the datafiles having ZTILELEN=100. I just included a single tile because of the size, is this ok or do we want 2 tiles (approx. 60MB)

MaxNoe:
Fine with me, I would go for two tiles.

MaxNoe:
Why is the zshrink file a gz?

tarrox:
>  Why is the zshrink file a gz?

Wrong name while piping the data out, fixed it

MaxNoe:
Could you live the sh script into a scripts directory?

tarrox:
I now pushed the testfiles and added them to the ZFitsReaderTests
I also created a scripts directory and moved testSubsetZfits.sh to it

src/main/java/fact/ZFitsTester.java
```diff
@@ -0,0 +1,103 @@
...
+import java.io.StringWriter;
+import java.net.MalformedURLException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ZFitsTester {
```
mackaiver: Missing javadoc

src/main/java/fact/io/hdureader/FITSStream.java
```diff
@@ -47,10 +47,21 @@ public void init() throws Exception {
         //get calibration constants they are stored in the first (and only) row of this hdu.
         HDU offsetHDU = fits.getHDU("ZDrsCellOffsets");
         if (offsetHDU != null){
-            ZFITSHeapReader reader = ZFITSHeapReader.forTable(offsetHDU.getBinTable());
-            boolean ignoreWrongTileHeader;
-            OptionalTypesMap<String, Serializable> row = reader.getNextRow(ignoreWrongTileHeader=true);
+            Boolean ztable = offsetHDU.header.getBoolean("ZTABLE").orElse(false);
```
mackaiver: is this for a case when the zdrscelloffsets are not compressed? does that ever happen?

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -130,17 +132,56 @@ public static ZFITSHeapReader forTable(BinTable binTable){
...
+     * @param ignoreWrongTileHeader if true, ignores wrong tileheader information.
      * @return a map containing the the data from the rows columns.
      * @throws NoSuchElementException iff hasNext() is false
      */
     public OptionalTypesMap<String, Serializable> getNextRow(boolean ignoreWrongTileHeader) throws IOException {
         if(!hasNext()){
            throw new NoSuchElementException();
         }
+        // check if we currently still have rows in the cache if not get the next tile
+        if (maxRowsInCache==curRowInCache) {
```
mackaiver: Since this iterator only walks forward a simple queue might suffice here right? No need to bookkeep three variables if you can use a datastructure that does the bookmarking automatically.

src/main/java/fact/io/hdureader/HDU.java
tarrox: its a widely used trick to calculate the ceil(a,b) of two ints faster, it is just that nowadays you rarely see it done, so people don't know about it as much, i added a comment explaining it a little bit

src/main/java/fact/io/hdureader/FITSStream.java
tarrox: yes in the same data as the one with ZTILELEN=100

src/main/java/fact/ZFitsTester.java
tarrox: done

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
tarrox: You would need a countdown type class, a queue would be a little bit of an overkill but yeah would work. Also, for this we only use two variables, so not much to gain anyways.

PR #268, opened 2017-11-28
Fix delay files

MaxNoe:
The Pixel Delay processor caught all exceptions and just logged them, which is why i did not catch the failing studies.


src/main/java/fact/datacorrection/CorrectPixelDelays.java
```diff
@@ -74,52 +74,49 @@ public void finish() throws Exception {
...
             log.error("Failed to load pixel delay data: {}", e.getMessage());
-            e.printStackTrace();
+            throw new RuntimeException(e);
         }
 
     }
-    
+
 
     /*
```
mackaiver: can we get rid of these?

PR #270, opened 2017-11-28
Add missing @Parameter in DrsCalibration, fixes #269

### LINKED ISSUE #269: missing @Parameter for outputKey of DrsCalibration
Result is "DataCalibrated" is hardcoded

MaxNoe:


tarrox:
This is also fixed in #246, in it it also adds other more parameters. If wanted i could extract all the processors that are changed that way and make an own pull request, if this would simplify things.

MaxNoe:
As this is a pretty small PR, I would propose to just merge this asap

PR #276, opened 2017-12-01
Write fits in standard analysis xmls, test correct number of events

MaxNoe:
* Remove a not working Test that was never executed (no @Test)


tarrox:
If this is a test for the analysis stuff, I think it would be nice to also check whether the written keys are correct (That all are written, and that they are the same between mc and data).

MaxNoe:
I think this could be added in another PR, for now, let's merge this as it already improves things quite dramatically.

PR #281, opened 2017-12-05
Implement access to version in fact-tools, print it on commandline

MaxNoe:
FACT-Tools now knows its version and git commit hash and reports it on the command line

The FITSWriter writes it to the FITS header of the outputfile

### DIFF
--- a/src/main/java/fact/VersionInformation.java
+++ b/src/main/java/fact/VersionInformation.java
@@ -1,0 +1,74 @@
+package fact;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+public class VersionInformation {
+
+    private final Logger log = LoggerFactory.getLogger(VersionInformation.class);
+
+    public final String version;
+    public final String commit;
+
+    private static VersionInformation instance;
+
+    public static VersionInformation getInstance() {
+        if (instance == null) {
+            instance = new VersionInformation();
+        }
+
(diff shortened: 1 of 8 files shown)


MaxNoe:
```
$ java -jar fact-tools-0.18.1-SNAPSHOT.jar
fact-tools, version v0.18.0-SNAPSHOT-96-g6f115e9

No container file specified.

Usage: 
	java -jar <fact-tools-jar> /path/container-file.xml

```

and 

```
$ java -jar target/fact-tools-0.18.1-SNAPSHOT.jar -v
project version: 0.18.1-SNAPSHOT
git description: v0.18.0-SNAPSHOT-96-g6f115e9
git commit hash: 6f115e9f5248e669de490965bd952faa7f903a2d
```

src/main/java/fact/VersionInformation.java
```diff
@@ -0,0 +1,98 @@
+package fact;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+public class VersionInformation {
```
tarrox: VersionInformation has no javadoc
Also do you realy want to go forth with the version stuff if it can't find a version at all?
I think breaking down would be a good thing as the version should be readable at all times.
I can't think of a reason why you couldn't have a version and still want to do stuff.

MaxNoe: The only case where it is not readable is when I run tests directly from intellij, I want to keep this functionality.

PR #283, opened 2017-12-07
Use mvn batch mode

MaxNoe:
Unclutters travis log


MaxNoe:
Did not achieve what I wanted, still clutters the log output

PR #290, opened 2017-12-15
Added test for checking outputkeys for null in the writers

tarrox:
I added the possibility to the JsonWriter and FitsWriter to test whether the given outputkeys are actually part of the data item. By default this is done, but can be disabled by setting the parameter allowNullKeys=true.

In addition I removed all the parameters that are not computed from the example analysis xml's. This were the *_sum parameters that are (I think) no longer computed by the ArrayStatistics Processor.
Also the MC files doesn't seem to contain the `@source` key so I removed it as well.


MaxNoe:
How does this handle the keys functionality of streams keys we use in some xmls,  e.g. using glob patterns and excludes?

```
keys="Theta*"
```

tarrox:
It ignores them, as in this cases the select method of Keys only selects keys that exists.

MaxNoe:
The code looks like it will throw the exception because it won't find them key `Theta*`

tarrox:
It could be interesting to remember the key-set of the previous process and fail if it changed. I will implement that also.

tarrox:
> The code looks like it will throw the exception because it won't find them key Theta*
Yes it will do that but I already fixed that just waiting for the test to pass.

src/main/java/fact/Utils.java
```diff
@@ -294,6 +295,41 @@ public static void mapContainsKeys(Data item, String... keys) {
 		}
 	}
 
+	/**
```
MaxNoe: Why reimplement this method?

tarrox: the orginal method needed String[], this allows List<String> it removes the need to convert a List into an array although I could change it into an collection instead of a list.

tarrox: changed it in the last to call the collection method if you want to use the String... keys method, so only one function is left

tarrox:
Fixed all the errors due to the changes and added the a test that checks whether the output-key-set changes from one event to the next. If this is the case I let it fail as this shouldn't happen.

MaxNoe:
Can you add the same check to the FITSWriter?

src/main/java/fact/io/FITSWriter.java
```diff
@@ -69,9 +73,29 @@
     @Override
     public Data process(Data item) {
         Data outputItem = DataFactory.create();
+        if (!allowNullKeys) {
```
MaxNoe: Maybe make this a method of a writer interface? So we avoid code duplication and can use it in the FITSWriter and the JSONWriter?

tarrox:
> Maybe make this a method of a writer interface? So we avoid code duplication and can use it in the FITSWriter and the JSONWriter?


Created an abstract class with the unified function. Maybe there is a nicer way to do this but I don't know one .

src/main/resources/default/settings_mc.properties
```diff
@@ -31,7 +31,8 @@ source_y=0.0
 
 # Outputkey Settings
 
-event_param=@source,EventNum
+#event_param=@source,EventNum
+event_param=EventNum
```
MaxNoe: Dont leave it it commented out, just remove it

tarrox: I just left it there for now, as i wasn't sure what the purpose of it was, as it seemed to be something important. But now that nobody is saying anything about it I assume it is not important, given that it is missing anyways. So I now removed it.

tarrox:
There are now a lot of theta_*_recPos keys not used anymore, I think someone refactored something and didn't update the settings.properties. I think we should merge this pull request faster than slower, so stuff like this gets seen by the person changing stuff and not me.

PS: removed the keys in the lastest commit.

MaxNoe:
> There are now a lot of theta_*_recPos keys not used anymore, I think someone refactored something and didn't update the settings.properties.

That has to be some kind of merging error, in the current master the `recPos` is gone and was substituted by `reconstructedSourcePositionX` and `reconstructedSourcePositionY`

PR #302, opened 2017-12-21
Fix caller lookup, fixes #301

### LINKED ISSUE #301: Wrapping mapContainsKey Broke calling processor in error message

MaxNoe:


MaxNoe:
Ok, I added the check for the method name and a test case.

PR #303, opened 2018-01-02
Sort items in viewer selector list

MaxNoe:
I was quite annoyed by the items not being ordered


mackaiver:
I prefer this. What do you think?

MaxNoe:
I like it, but you have to approve.

PR #305, opened 2018-01-03
Skip 27s events, fixes #150

### LINKED ISSUE #150: Flag 27sec events
We (Zurich) think the 27sec events should be skipped entirely during analysis.

An event is a 27sec event if many pixels are saturated for > 100 slices.
(we assume, if one pixel saturates for >100 slices ... then all do ... so maybe checking one pixel is already enought... but more are better.

We think flagging these files as "Is27SecondEvent" would be good.

We further think in Jens `CorrectSaturation` processor is well suited to find these.

MaxNoe:
This implements skipping of the 27s events in the standard analysis and photonstream extraction xmls.

While I did the implementation I looked into the file `20151218_161`, as this has a lot of 27s early in the file (first is `EventNum` 12).

I noticed that these events take considerably longer to analyse than even events with large showers, in the order of multiple seconds, so I went for skipping them early instead of flagging them.

The typical 27s second event in that file had round about 60 % saturated slices. I put the skip threshold to 40 %. I think this is quite save as even if a physics shower could produce 40 % saturation it 1. would be not reliable to reconstruct and 2. be very likely a hadron event.


MaxNoe:
I think we should have a little discussion about this, so I'll wait for the others to comment.

MaxNoe:
Here is a histogram of the `saturatedSlicesPercentage` for a single run with 27s events (only events where  `saturatedSlicesPercentage` > 0. All Events below 0.2 (maximum is 0.05) look like bright showers, all above look like 27s events.

saturation_27s

PR #308, opened 2018-01-09
Fix alle unchecked, unsafe or deprecetion warnings

MaxNoe:


src/main/java/fact/features/muon/HoughTransform.java
```diff
@@ -91,7 +91,7 @@
     private double[] circle_x;
     private double[] circle_r;
 
-    public ArrayList<int[]>[] chid2circles = new ArrayList[Constants.NUMBEROFPIXEL];
+    public ArrayList<ArrayList<int[]>> chid2circles = new ArrayList<>();
```
mackaiver: You could initialize the arraylists to the proper length(if you know it). Makes it somewhat more efficient.

PR #318, opened 2018-01-18
Single analysis

MaxNoe:
* All modules now as single version for data and mc.

* Only one `settings.properties` file

* Global options for Processors are now named `<processor>.<option>`, e.g. `TwoLevelTimeNeighbor.coreThreshold=5`

* Two set of predefined output keys: `outputKeysObservations`, `outputKeysSimulations`

* Addionaly if source dependent features should be calculated: `source_dependent_observations` and `...simulations`

* Move skips from the included xmls to the including xmls to be more explicit

Unfortunately, the including xmls need to be somewhat different, setting different outputKeys for example.
And auxService and pixelDelayFile no also needs to be set for mcs.

Compare `examples/stdAnalysis/observations.xml` with `examples/studies/simulations.xml`


examples/apply_model.xml
```diff
@@ -21,49 +21,25 @@
...
+        <stream.flow.Skip condition="%{data.TriggerType} != 4" />
+
+        <include url="classpath:/analysis/calibration.xml" />
+        <include url="classpath:/analysis/extraction.xml" />
+        <include url="classpath:/analysis/cleaning.xml" />
+        <include url="classpath:/analysis/pedestalParameters.xml" />
+        <include url="classpath:/analysis/imageParameters.xml" />
+        <include url="classpath:/analysis/sourcePosition.xml" />
 
         <fact.ApplyModel url="file:/home/kai/Development/erna/test.pmml" />
```
mackaiver: we should change this

MaxNoe: Will do so in another PR providing a test pmml for unit testing.

MaxNoe:
@mackaiver @jebuss I think this is ready to be merged now.

MaxNoe:
Ok, now we have only one file for the standard analysis, this was achieved by making the Writers `keys` parameter optional and then getting the correct output keys from the `settings.properties` file.

PR #327, opened 2018-02-15
Change coordinate system

MaxNoe:
We rotate the coordinate system by 90 degrees counter-clockwise
so that when one looks onto the camera from the telescope dish,
x points right and y points up.

This is the same coordinate system used by MARS and makes
it easier to plot things into the camera, as x and y
fit intuition.

This is a breaking change that requires changes in all
downstream software handling coordinate transformations (e.g. pyfact)


MaxNoe:
To check that we did not break the Coordinate transformation, I ran this version on the star pedestals we took in 20151115, you can nicely see how the by FACT-Tools calculated star positions in the camera coincide with higher values of pedvar:

aldebaran
algol

mackaiver:
> This is a breaking change that requires changes in all
> downstream software handling coordinate transformations (e.g. pyfact)

Does that mean we can't analyze old data anymore?  How do you propose we handle that? Can we add version information to the data files?

MaxNoe:
> Does that mean we can't analyze old data anymore? How do you propose we handle that? Can we add version information to the data files?

With pyfact / classifier-tools versions newer than the version that incorporates this change, we would not be able to analyse old data, if we do not make the rotation optional.
We could however do that with a command line switch for the tools that use it (`--old-coordinate-system` or `--no-coord-rotation`

MaxNoe:
I think using a more intuitive coordinate system that is the same as MARS makes up for this inconvinience

PR #328, opened 2018-02-16
Support classpath:/ urls for input files

MaxNoe:
This fixes some long standing issues of using `-Ddrsfile=classpath:/testDrsFile.fits.gz`  or something similar

### DIFF
--- a/src/main/java/fact/ClasspathURLStreamHandlerFactory.java
+++ b/src/main/java/fact/ClasspathURLStreamHandlerFactory.java
@@ -1,0 +1,29 @@
+package fact;
+
+
+import java.io.IOException;
+import java.net.*;
+
+public class ClasspathURLStreamHandlerFactory implements URLStreamHandlerFactory {
+
+    @Override
+    public URLStreamHandler createURLStreamHandler(String protocol) {
+        if ("classpath".equals(protocol)) {
+            return new ClasspathURLStreamHandler();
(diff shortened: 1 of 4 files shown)


mackaiver:
This is nice. Does that mean we can remove this `SourceURL` business everywhere?

MaxNoe:
If I didn't overlook anything, the only place where it is still used is `Stream` implementations,  as the constructor requires a`SourceURL`

mackaiver:
So this means we can finally add classpathes to the xmls here right?
https://github.com/fact-project/fact-tools/blob/ec792dd0ea13d0d880438e171493d3bea5b4e85d/examples/stdAnalysis.xml#L4-L6

MaxNoe:
Oh yes, this should work now!

MaxNoe:
For the inputfiles, it works but not for the aux folder

MaxNoe:
The `AuxFileService` uses `Path`s, I think these cannot handle classpath resources

mackaiver:
To make this work with auxfile thingy I'd have to restructure the HDUreader to accept inputstreams. That's not entirely trivial. Maybe another time.

mackaiver:
Refactored a bit  in the last two commits.

mackaiver:
Wow. I can approve my own commits. Nice. 🥇

PR #332, opened 2018-02-16
Fix hdureader for ommitted repeat count, fixes #331

### LINKED ISSUE #331: HDUReader does not respect missing repeat count
The fits standard states:

> t he repeat count r is the ASCII representation of a non-negative integer specifying the number of elements in field n. The default value of r is 1; the repeat count need not be present if it has the default value

This is not handled in the hdureader and it is not able to read files  correctly that use this.

MaxNoe:
This happend for a file produced by astropy, wonderwhy this never happend before


mackaiver:
Nice. I remember fixing this at one point. Apparently I never pushed it.

PR #333, opened 2018-02-16
Gain service

### LINKED ISSUE #267: Consider changing gains in Extraction
Individual Pixels' gains do change over time. While some mail functions of pixels are considered in the Calibration Service, the variation of the gains is not.

In the worst case, the median camera gain varies by (30 mV slices) and the minimum gain even by (250 mV slices), in both cases this was determined on a window size of 30 slices and the generation of single pe spectra.

Here you see the time series of the gain measured in single pe spectras
image

This is the distribution of the median values of the time series above.
image

This is the distribution of the minima of the time series above.
image

I suggest to check if these variations have to be considered and if this is the case, a gain service for the extraction algorithm.

MaxNoe:
Fixes #267 

* Introduces a new service which reads in a fits file with the gains of the single pe measurements
* Fits file in the resources with values from 20130825 -- 20170917
* If `UnixTimeUTC` is not in the date item, the defaultIntegralGains are used
* warning if time difference between event and gain measurement is larger than 5 days

Question: can we create a file with a larger time range?


src/main/java/fact/gainservice/GainService.java
```diff
@@ -0,0 +1,146 @@
...
+        return gain;
+    }
+
+
+    public double[] getSimulationGains() {
+        if (gainsSimulations == null) {
+            loadGainsSimulations();
+        }
+        return gainsSimulations;
+    }
```
jebuss: Is this getter really needed since it only calls loadGainsSimulations

MaxNoe:
This is way I added the `InterpolatePixelArray` processor to the unit tests, the HillasParameterisation failed because photoncharge was `-Infinity`. Maybe we should just interpolate pixels with `gain == 0` instead or on top of using the hardware service. But at least for the test data the broken pixels list worked fine.

src/main/java/fact/gainservice/GainService.java
MaxNoe: ähm. I think so.

It does not only call `loadGainsSimulation`. It only executes `loadGainsSimulations` if it hasn't already done it. So the getter is necessary to not always read the csv file but only read it once.

MaxNoe:
Should we merge this now? Or should we wait until we can provide a file with larger time range? I would rather not put too many largish binary files in the history

PR #335, opened 2018-03-02
Use default gains

MaxNoe:
The results on the open data sample got worse with the new gain service.

This PR enbales using our old "default" gains with the new gain service and makes it the defautl


MaxNoe:
@fact-project/fact-tools-core-dev I wouldn't merge this yet but first look into #336

MaxNoe:
superseeded by #336

PR #336, opened 2018-03-05
Add new default gain file for MCs simulated with ceres 12

jebuss:
Rerunning the single pe analysis on simulated closed shutter pedestal files from ceres_12 revealed that the default gains used in `fact.src.resources.defaultIntegralGains.csv` (242 mV 0.5 ns) are not correct anymore.

This adds a calibration file with gains according to a single pe analysis on pedestal simulations with ceres_12. 
Entries are the median of all 1440 pixels (except for the 12 broken pixels), rounded to zero digits. 

The result of the spe analysis was:

mean/median = 266.0 mV 0.5 ns
std = 8.0 mV 0.5 ns


jebuss:
@MaxNoe I guess we also want to change the path in the gain service, so I modified it accordingly.

jebuss:
@MaxNoe please have a look. Anything else missing?

MaxNoe:
Tests are failing

jebuss:
> I think we should keep them side by side for a while, until we could show that things got better with the new gain file. Right now, old mc gain file and old data gain file is the combination with the smallest mismatch. Maybe it would even be good if you can choose the file for the gain service.  However, the old file can be deleted once improvement has been proven.

PR #337, opened 2018-03-07
Update cleaning levels

MaxNoe:
Lower cleaning levels to adapt to lower sizes as result of the GainService and new MC Gain file.

Significance on the open crab sample is now 25 σ:
theta2_c5025


jebuss:
How large was LiMa before?

jebuss:
Verrrry Nice!

MaxNoe:
> How large was LiMa before?

* With new GainService but old MC Gain 21 σ
* With old default gains and old MC gain 23.8 σ

PR #341, opened 2018-03-09
Set skipErrors to `false` for erna processing

MaxNoe:


MaxNoe:
As this behaviour is now default, it's not needed anymore

PR #348, opened 2018-03-16
Parameter improvment and some small new abilities

tarrox:
* Improvement to the current processors by adding additional parameters to set,
primarily adding the ability to set the UnixTimeUTC key.
* Added the ability to reverse the process for the DrsCalibration and the remapping.
* Added a processor to combine two arrays into one by combining each element, given the four std operators.

This changes are necessary for the work I do, to make the pull easier I split the last pull request #246 up into three pull request this being the first one.


src/main/java/fact/datacorrection/InterpolateTimeSeries.java
```diff
@@ -36,6 +36,12 @@
     @Parameter(required = false, description = "The minimum number of neighboring pixels required for interpolation", defaultValue = "3")
     public int minPixelToInterpolate = 3;
 
+    @Parameter(required = false, description = "The key for the resulting badPixelSet.")
+    public String badPixelKey = "badPixels";
+
+    @Parameter(required = false, description = "The key to the unixTimeUTC of the Event.")
+    public String unixTimeKey = "UnixTimeUTC";
```
MaxNoe: Why is this a parameter?

src/main/java/fact/datacorrection/InterpolateTimeSeries.java
```diff
@@ -36,6 +36,12 @@
     @Parameter(required = false, description = "The minimum number of neighboring pixels required for interpolation", defaultValue = "3")
     public int minPixelToInterpolate = 3;
 
+    @Parameter(required = false, description = "The key for the resulting badPixelSet.")
+    public String badPixelKey = "badPixels";
```
MaxNoe: Why is this an xml Parameter?

src/main/java/fact/datacorrection/PatchJumpRemoval.java
```diff
@@ -38,6 +38,9 @@
     @Parameter(required = true)
     public String startCellKey = null;
 
+    @Parameter(required = false, description = "The key containing the UnixTimeUTC")
+    public String unixTimeKey = "UnixTimeUTC";
```
MaxNoe: Why is this a parameter?

src/main/java/fact/features/source/SourcePosition.java
```diff
@@ -58,6 +58,9 @@
     @Service(required = false, description = "Name of the service that provides aux files")
     public AuxiliaryService auxService;
 
+    @Parameter(required = false, description = "The key containing the UnixTimeUTC")
+    public String unixTimeKey = "UnixTimeUTC";
```
MaxNoe: Why is this a Parameter?

src/main/java/fact/utils/CombineDataArrays.java
```diff
@@ -0,0 +1,85 @@
...
+            throw new RuntimeException("The given operation op: " + op + " is not supported");
+        }
+
+        input.put(outputKey, resultArray);
+
+        return input;
+    }
+
+
+    public static Logger getLog() {
```
MaxNoe: Not needed

src/main/java/fact/utils/CombineDataArrays.java
```diff
@@ -0,0 +1,85 @@
...
+        return log;
+    }
+
+
+    public static void setLog(Logger log) {
```
MaxNoe: Not needed

src/main/java/fact/utils/PreviousEventInfo.java
```diff
@@ -18,7 +18,11 @@
     @Parameter(required = true)
     public String outputKey = null;
 
-    int limitEvents = 20;
+    @Parameter(required = false, description = "The key containing the UnixTimeUTC")
```
MaxNoe: Why is this a parameter?

MaxNoe:
Please also add `-SNAPSHOT` to the version in the pom

src/main/java/fact/datacorrection/InterpolateTimeSeries.java
tarrox: I use this processor twice for different inputs with different times one will use 'UnixTimeUTC' the other 'LONS_UnixTimeUTC'. This is due to the fact that i have to calibrate the pedestal that I superimpose.

src/main/java/fact/datacorrection/InterpolateTimeSeries.java
tarrox: I use the processor twice, so i added this in case someone want to preserve the output and not override the key with the second processor. I think it is cleaner this way, but I can remove it.

src/main/java/fact/datacorrection/PatchJumpRemoval.java
tarrox: I use this processor twice for different inputs with different times one will use 'UnixTimeUTC' the other 'LONS_UnixTimeUTC'. This is due to the fact that i have to calibrate the pedestal that I superimpose.

src/main/java/fact/features/source/SourcePosition.java
tarrox: I use this processor twice for different inputs with different times one will use 'UnixTimeUTC' the other 'LONS_UnixTimeUTC'. This is due to the fact that i have to get the sourceposition of the pedestal that I superimpose.

src/main/java/fact/utils/CombineDataArrays.java
tarrox: done

src/main/java/fact/utils/CombineDataArrays.java
tarrox: done

src/main/java/fact/utils/PreviousEventInfo.java
tarrox: For the case someone wants to set it, we have an variable in 'settings.properties' for this but no way to set it in this processor (only in the jumpRemoval). So either this one goes or, the thing in setting.properties makes little sense because it is a constant.

tarrox:
Added the -SNAPSHOT to the version.

src/main/java/fact/features/source/SourcePosition.java
MaxNoe: What is LONS supposed to mean?

MaxNoe:
As 1.0.1 is released now, the version for this should be `1.0.2-SNAPSHOT`

src/main/java/fact/features/source/SourcePosition.java
tarrox: 'Light of night sky' before it was PED but we already have stuff that is called pedestal so I use LONS, that way there is no conflict with the other stuff.

MaxNoe:
As we now require the timestamp key anyways, we could remove  conversions from unixtimeutc to zoneddatetime and just do it once, maybe even in the readers and just use the timestamp key everywhere

jebuss:
> As we now require the timestamp key anyways, we could remove conversions from unixtimeutc to zoneddatetime and just do it once, maybe even in the readers and just use the timestamp key everywhere

I'm not sure if I know what you mean. Could you point me to a file / line?

tarrox:
> I'm not sure if I know what you mean. Could you point me to a file / line?

I think he means this neat feature here `<fact.features.UnixTimeUTC2DateTime />` which does the conversation once at the beginning. This would remove the necessity to convert it be hand in every processor like in  `datacorrection/InterpolateTimeSeries.java:52`.
I would agree with Max point that this would indeed be better but I would suggest to leave it at this for now and create a different branch where every processor gets updated to the timestamp version.

MaxNoe:
> I'm not sure if I know what you mean. Could you point me to a file / line?

@jebuss Everywhere we do calculations involving timestamps we use `ZonedDateTime`. We always convert `UnixTimeUTC` into timestamp. We could just do this once, maybe even in the readers instead of doing it in each processor that needs to do stuff with a timestamp.

We use `Utils.unixTimeUTCToZonedDatetime` in these processors

```
src/main/java/fact/cleaning/TwoLevelTimeMedian.java
src/main/java/fact/cleaning/TwoLevelTimeNeighbor.java
src/main/java/fact/features/UnixTimeUTC2DateTime.java
src/main/java/fact/features/source/CameraToEquatorial.java
src/main/java/fact/features/source/SourcePosition.java
src/main/java/fact/datacorrection/InterpolateTimeSeries.java
src/main/java/fact/datacorrection/InterpolatePixelArray.java
src/main/java/fact/extraction/BasicExtraction.java
src/test/java/fact/io/hdureader/BinTableTests.java
```

jebuss:
Ok, @MaxNoe I see your point. IMHO, it is neither a show stopper for this PR nor is there a larger clarity in opening a new issue and a new PR for this, since this PR is anyways a collection of several features. So both ways do make sense. 

So I would say its up to @tarrox if he can put afford in adding the `timestamp` in the readers and adapt the according processors for this PR. It does not seem to be much of an afford. However, If he does not have time right now, I would also vote for opening an issue and and a new PR for this feature.

I would be happy if we could finish this PR soon.

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -108,52 +119,67 @@ public Data process(Data data) {
...
 
-        double[] calibrated = applyDrsCalibration(rawfloatData, output, startCell);
-
-        data.put(outputKey, calibrated);
+        double[] calibrated = applyDrsCalibration(rawfloatData, output, startCell, reverse);
+        if (!reverse)
+            data.put(outputKey, calibrated);
+        else {
+            short[] decalibratedShortData = new short[calibrated.length];
+            // System.arraycopy(rawData, 0, rawfloatData, 0, rawfloatData.length);
```
MaxNoe: Commented code

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -108,52 +119,67 @@ public Data process(Data data) {
...
+            for (int i = 0; i < calibrated.length; i++) {
+                decalibratedShortData[i] = (short) calibrated[i];
+            }
+            data.put(outputKey, decalibratedShortData);
+        }
 
-        // add color value if set
+        //TODO add color value if set
```
MaxNoe: Please just remove this line

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -240,13 +266,24 @@ public Data process(Data data) {
...
 
-                triggerOffsetPos = pixel * drsTriggerOffsetMean.length / Constants.N_PIXELS + slice;
-
-                vraw = data[pos] * dconv;
-                vraw -= drsBaselineMean[offsetPos];
-                vraw -= drsTriggerOffsetMean[triggerOffsetPos];
-                vraw /= drsGainMean[offsetPos];
-                vraw *= 1907.35;
+                triggerOffsetPos = pixel * drsTriggerOffsetMean.length / Constants.N_PIXELS
+                        + slice;
```
MaxNoe: this line break makes things worse imho

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -108,52 +119,67 @@ public Data process(Data data) {
...
+            data.put(outputKey, decalibratedShortData);
+        }
 
-        // add color value if set
+        //TODO add color value if set
 
         return data;
     }
 
     public double[] applyDrsCalibration(double[] data, double[] destination,
```
MaxNoe: I still think two functions would be more clear. `applyDrsCalibration` and `revertDrsCalibration`.

src/main/java/fact/datacorrection/DrsCalibration.java
tarrox: Removed this one and the other.

src/main/java/fact/datacorrection/DrsCalibration.java
tarrox: removed

src/main/java/fact/datacorrection/DrsCalibration.java
tarrox: fixed this one and the other ones too

src/main/java/fact/datacorrection/DrsCalibration.java
tarrox: split into two functions

tarrox:
Replaced the `UnixTimeUTC` with `timestamp` in the processors where it makes sense.

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -264,6 +289,47 @@ public Data process(Data data) {
...
+    /**
+     * Reverses the drsCalibration performed in applyDrsCalibration.
+     *
+     * @param data The calibrated data the decalibrate.
+     * @param destination If given use this as the destination array otherwise a new one is created.
+     * @param startCellVector The array containing the start cells used to know which calibration constants to use.
+     * @return The decalibrated data array
+     */
+    public double[] reverseDrsCalibration(double[] data, double[] destination, short[] startCellVector) {
+        if (destination == null || destination.length != data.length)
```
MaxNoe: Maybe throwing in case the length do not match is better, what do you think?

src/main/java/fact/datacorrection/PatchJumpRemoval.java
```diff
@@ -94,10 +97,10 @@ public Data process(Data item) {
         Utils.isKeyValid(item, prevEventsKey, PreviousEventInfoContainer.class);
         Utils.isKeyValid(item, startCellKey, short[].class);
         Utils.isKeyValid(item, "NROI", Integer.class);
-        Utils.isKeyValid(item, "UnixTimeUTC", int[].class);
+        Utils.isKeyValid(item, unixTimeKey, int[].class);
 
         // Get variables out of data item
-        int[] currentTime = (int[]) item.get("UnixTimeUTC");
```
MaxNoe: This processor apperently still uses `UnixTimeUTC` and not timestamp

tarrox:
Added issue #362 for the missing unixtime->timestamp changes. So we can finish this PR quickly.

PR #349, opened 2018-03-16
Hdu reader skip rows

tarrox:
Added the ability to skip rows in the HDU reader. This is nedded for the SamplePedestalEvent processor that is coming in a later pull request.

This is the second part of the pull request simplification of #246 and can be merged independently of #348.

### DIFF
--- a/src/main/java/fact/io/hdureader/BinTable.java
+++ b/src/main/java/fact/io/hdureader/BinTable.java
@@ -35,6 +35,12 @@ public class BinTable {
     final DataInputStream tableDataStream;
     DataInputStream heapDataStream = null;
 
+    private final Header header;
+
+    public Header getHeader() {
+        return header;
+    }
+
 
     /**
      * This enum maps the type characters in the header to the fits types.
@@ -130,8 +136,10 @@ public class BinTable {
     public Integer numberOfRowsInTable = 0;
     public Integer numberOfColumnsInTable = 0;
     public final String name;
+    public Integer numberOfBytesPerRow = 0; // in header value: naxis1
 
     BinTable(Header header, long hduOffset, URL url) throws IllegalArgumentException, IOException {
+        this.header = header;
 
         binTableSanityCheck(header);
 
@@ -163,6 +171,7 @@ public class BinTable {
             columns.add(new TableColumn(zform, tform, ttype.getValue()));
         }
 
+        numberOfBytesPerRow = header.getInt("ZNAXIS1").orElse(header.getInt("NAXIS1").orElse(0));
         numberOfRowsInTable = header.getInt("ZNAXIS2").orElse(header.getInt("NAXIS2").orElse(0));
         numberOfColumnsInTable = columns.size();
 
--- a/src/main/java/fact/io/hdureader/BinTable.java
+++ b/src/main/java/fact/io/hdureader/BinTable.java
@@ -35,12 +35,7 @@ public class BinTable {
     final DataInputStream tableDataStream;
     DataInputStream heapDataStream = null;
 
-    private final Header header;
-
-    public Header getHeader() {
-        return header;
-    }
-
+    public final Header header;
 
     /**
      * This enum maps the type characters in the header to the fits types.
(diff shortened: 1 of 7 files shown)


src/main/java/fact/io/hdureader/BinTable.java
```diff
@@ -35,6 +35,12 @@
     final DataInputStream tableDataStream;
     DataInputStream heapDataStream = null;
 
+    private final Header header;
```
MaxNoe: why not public final?

src/main/java/fact/io/hdureader/FITSStream.java
```diff
@@ -127,4 +127,8 @@ public Data readNext() throws Exception {
 
         return item;
     }
+
+    public Reader getReader() {
```
MaxNoe: Is this just for tests? Then Reader could be package private and the getter is unnecessary

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -106,6 +112,29 @@ public static ZFITSHeapReader forTable(BinTable binTable) {
         return new ZFITSHeapReader(binTable);
     }
 
+    @Override
+    public void skipToRow(int num) throws IOException {
+        //calc zshrink problem
+        //num = num+1;
```
MaxNoe: remove or explain

src/main/java/fact/io/hdureader/FITSStream.java
tarrox: I need this also later in the `SamplePedestal` processor but I changed it to a skipToRow function as this is the only thing I need. No reason to expose the whole reader.

src/main/java/fact/io/hdureader/BinTable.java
tarrox: I learned not to expose any variables outside so I did that here, but yeah it is final it can be public. Changed it and removed the getter.

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
tarrox: fixed the error and thus the problem with zshrink

src/main/java/fact/io/hdureader/BinTableReader.java
```diff
@@ -141,4 +143,13 @@ private Serializable readArrayFromStream(BinTable.TableColumn c, DataInputStream
         }
         return null;
     }
+
+    @Override
+    public void skipToRow(int num) throws IOException {
```
mackaiver: Can you write a little docstring here?

src/main/java/fact/io/hdureader/FITSStream.java
```diff
@@ -127,4 +127,8 @@ public Data readNext() throws Exception {
 
         return item;
     }
+
+    public void skipToRow(int num) throws IOException{
```
mackaiver: Some javadoc would be nice

src/main/java/fact/io/hdureader/Reader.java
```diff
@@ -72,5 +72,12 @@
      */
     OptionalTypesMap<String, Serializable> getNextRow() throws IOException;
 
+    /**
+     * Skips the given number of rows.
+     *
+     * @param num The amount of rows to skip.
+     * @throws IOException
+     */
+    void skipToRow(int num) throws IOException;
```
mackaiver: It skips 'num' rows right? So if i call 'skipToRow(4)' twice in a row im looking at row 8 right? I propose to change the name of the method to simply 'skipRows'

src/main/java/fact/io/hdureader/BinTableReader.java
tarrox: sure and done

src/main/java/fact/io/hdureader/FITSStream.java
tarrox: Done

src/main/java/fact/io/hdureader/Reader.java
tarrox: It skipped to the row from the beginning and could only be called from there, I changed it now to be able to skip rows whenever

tarrox:
As said in the reply to @mackaiver, the skipToRow function is now a skipRows function that allows to skip from any position in the file.

tarrox:
Removed the commented lines. If this was all it should be ready for merge.

PR #350, opened 2018-04-02
Update .zenodo.json

MaxNoe:


MaxNoe:
This works now: https://zenodo.org/record/1211604/

I will backport this to 0.17.2 and 1.0.0 so we have dois for these versions we can cite

Issue #353, opened 2018-04-30
std analysis erna observations does not work anymore with erna.process_fact_data


jebuss:
Using the recent fact-tools version with erna.process_fact_data does not work anymore for two reasons:

1. it still uses the old `fact.auxservice.SqliteService`, which is not needed anymore since there is `fact.auxservice.AuxFileService` that can handle directory trees.

https://github.com/fact-project/fact-tools/blob/328feab9e6b9136057e410a8227395a206298b6e/examples/forErna/std_analysis_erna_observations.xml#L8-L12

2. erna provides the drs-file as key-value-pair in the data item via `fact.io.FactFileListMultiStream` but `classpath:/analysis/calibration.xml` uses `fact.datacorrection.DrsCalibration` with a URL given by a property `"${drsfile}"` this way the key `@drsFile`is never checked. 

https://github.com/fact-project/fact-tools/blob/328feab9e6b9136057e410a8227395a206298b6e/src/main/resources/analysis/calibration.xml#L11-L15

I suggest:

1. exchange `fact.auxservice.SqliteService` by `fact.auxservice.AuxFileService`
2. delete: https://github.com/fact-project/fact-tools/blob/328feab9e6b9136057e410a8227395a206298b6e/src/main/resources/analysis/calibration.xml#L12
and put a `<stream.data.SetValue key="@drsFile" value "${drsfile}"/>` into the std-analysis.

tarrox:
This problem is also in part a problem that is also discussed in #351

Issue #351, opened 2018-04-09
Make calibration.xml work with FactFileMultistream


tarrox:
Currently the `DrsCalibration` in the `calibration.xml` uses the `url="${drsfile}"` method of getting the drs file, which makes the xml incompatible with the use of `FactFileMultistream`. I think it would be nice if it would work with it, as it means one can just reuse it.

There are two options I can think of to make this work:

* remove the `ulr=${drsfile}` and have a processor that injects the `${drsfile}` into the datastream as the `@drsFile` item.
* by default or by parameter, make it that the `DrsCalibration`-Processor switches to the `@drsFile` option if the given url can't be read (For example because the string is empty or because it couldn't be read).

MaxNoe:
Maybe something like:

```java
if (item.get("@drsFile") != null && url == "") {
    // use @drsFile
} else {
    // use url
}
```

MaxNoe:
So streams does apparently not allow to create an empty or null url by setting the property to an empty string.

What I came up with was setting `-Ddrsfile=file:` and using:

```java
    @Override
    public void init(ProcessContext processContext) throws Exception {

        if (url != null)  {
            if (url.getPath() != "") {
                try {
                    loadDrsData(url);
                } catch (Exception e) {
                    log.error("Could not load .drs file specified in the url.");
                    throw new RuntimeException(e.getMessage());
                }
            } else {
                url = null;
            }
        }
    }
```

But I don't really like it. Anybody got a better idea?

tarrox:
Another option would be to switch the url to a string and convert it if needed. That way we could set it as empty. Although this wouldn't be a good solution either, i would even say worse.

Another way would be to skip the whole init function and simply load the url if needed in the process function, basically taking `@drsfile` and if not given use the url parameter, if this fails, just abort as usual. That way we wouldn't need to check for an empty path as a way to check which method to use. Instead the drskey is always used if given.
I think this option would be the cleanest.

MaxNoe:
> simply load the url if needed in the process function

The checking is done by streams at the xml Parsing level. The only thing would be make it a string and convert to url.

PR #358, opened 2018-05-09
Use AuxFileService in examples/forErna/std_analysis_erna_observations.xml

jebuss:
The `SqliteService` is not needed anymore since the `AuxFileService` is able to go through the entire directory tree


MaxNoe:
We could also completely remove the `SqliteService`

jebuss:
this will fix part 1 of #353

jebuss:
fixes #353

jebuss:
fixes #351

PR #359, opened 2018-05-09
DrsFileService

mackaiver:
Adds a DrsFileService which looks for the proper `.drs` file for the currently analysed raw data file.

### DIFF
--- a/src/main/java/fact/datacorrection/DrsCalibration.java
+++ b/src/main/java/fact/datacorrection/DrsCalibration.java
@@ -4,6 +4,7 @@
 package fact.datacorrection;
 
 import fact.Constants;
+import fact.DrsFileService;
 import fact.io.hdureader.BinTable;
 import fact.io.hdureader.BinTableReader;
 import fact.io.hdureader.FITS;
@@ -14,6 +15,7 @@ import stream.Data;
 import stream.ProcessContext;
 import stream.StatefulProcessor;
 import stream.annotations.Parameter;
+import stream.annotations.Service;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


jebuss:
I think it would be nice if the used drs file (ID/filename) is part of the data item and also of the output of fact-tools

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -277,15 +297,16 @@ public void init(ProcessContext processContext) throws Exception {
         }
```
jebuss: can you add the drs_url to the data item in the process method? Or a key value par containing night and run id of the drs file.

jebuss: This way it is more reproducable

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -34,15 +36,20 @@
...
-    @Parameter(required = false, description = "Data array to be calibrated", defaultValue = "Data")
-    public String key = "Data";
+	@Parameter(required = false, description = "Data array to be calibrated", defaultValue = "Data")
+	private String key = "Data";
 
-    @Parameter(required = false, description = "A URL to the DRS calibration data (in FITS formats)",
-            defaultValue = "Null. Will try to find path to drsFile from the stream.")
-    public URL url = null;
+	@Parameter(required =  false, description = "A URL to the DRS calibration data (in FITS formats)",
+			defaultValue = "Null. Will try to find path to drsFile from the stream.")
```
tarrox: Add `or through the drsservice is given.`

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -96,8 +103,21 @@ protected void loadDrsData(URL in) {
...
+				this.drsGainMean = calibrationInfo.drsGainMean;
+				this.drsTriggerOffsetMean = calibrationInfo.drsTriggerOffsetMean;
 
-        if (this.url == null) {
+			} catch (NullPointerException | IOException e) {
+				throw new RuntimeException("Could not get calibration info from service. \n" + e.getMessage());
+			}
+		}
+
+        if (this.url == null && this.drsService == null) {
```
tarrox: Change `if` to `else if`

PR #360, opened 2018-05-09
Empty drs file, fixes #353, fixes #351

### LINKED ISSUE #351: Make calibration.xml work with FactFileMultistream

### LINKED ISSUE #353: std analysis erna observations does not work anymore with erna.process_fact_data

MaxNoe:


jebuss:
fixes #353

jebuss:
fixes #351

MaxNoe:
Yes, the idea was to set it to the empty string!

MaxNoe:
It has to be set, but now the empty string is a valid value.

jebuss:
Make sense to me, but even with en empty property ` <property name="drsfile" value="" />`  I still get:

```
$ java -jar target/fact-tools-v1.0.2-12-g0bd380a-dirty.jar examples/forErna/std_analysis_erna_observations.xml
streams exited on error: Error while creating instance of 'fact.datacorrection.DrsCalibration'
java.lang.IllegalArgumentException: Error while creating instance of 'fact.datacorrection.DrsCalibration'
	at stream.runtime.setup.factory.DefaultProcessFactory.createProcessorAndRegisterServices(DefaultProcessFactory.java:295)
	at stream.runtime.setup.factory.DefaultProcessFactory.createNestedProcessors(DefaultProcessFactory.java:277)
	at stream.runtime.setup.factory.DefaultProcessFactory.createAndRegisterProcesses(DefaultProcessFactory.java:255)
	at stream.runtime.setup.handler.ProcessElementHandler.handleElement(ProcessElementHandler.java:96)
	at stream.runtime.ProcessContainer.init(ProcessContainer.java:517)
	at stream.runtime.ProcessContainer.<init>(ProcessContainer.java:401)
	at stream.runtime.ProcessContainer.<init>(ProcessContainer.java:250)
	at stream.run.main(run.java:164)
	at stream.run.main(run.java:147)
	at fact.run.main(run.java:26)
Caused by: java.lang.IllegalArgumentException: Property drsfile is not set.
	at stream.util.Variables.substitute(Variables.java:157)
	at stream.util.Variables.substitute(Variables.java:121)
	at stream.util.Variables.expand(Variables.java:113)
	at stream.runtime.setup.factory.ObjectFactory.create(ObjectFactory.java:243)
	at stream.runtime.setup.factory.ObjectFactory.create(ObjectFactory.java:214)
	at stream.runtime.setup.factory.DefaultProcessFactory.createProcessorAndRegisterServices(DefaultProcessFactory.java:293)
```

Dunno why that is

src/main/java/fact/datacorrection/DrsCalibration.java
```diff
@@ -267,11 +269,20 @@ public Data process(Data data) {
 
     @Override
     public void init(ProcessContext processContext) throws Exception {
-        if (url != null) {
+
+        if (!url.isEmpty()) {
```
jebuss: it seems that in case the property is empty this line is never reached. I assume that streams does not handle empty properties very well.

jebuss: However, searching for a specific sting works `if (!url.equals("a"))` and in the XML `<property name="drsfile" value="a" />`.

MaxNoe:
I opened a PR in streams to enable setting empty properties here: https://bitbucket.org/cbockermann/streams/pull-requests/15/allow-empty-properties

MaxNoe:
Just waiting for the tests to pass

PR #364, opened 2018-05-29
Sample pedestal noise

tarrox:
Adds the SamplePedestalEvent processor to load a given pedestal event in as noise.

Also added two xml files.
 * `add_pedestal_noise_template.xml` adds the pedestal noise to mc events and saves them as xmls
 * `examples/pedestal_noise_mc_analysis.xml` analysis the created mc events like the normal analysis


examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,114 @@
+<container>
```
jebuss: could you add a doc string that is telling what the process' purpose and what it is doing?

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,114 @@
...
+        <fact.datacorrection.RemoveSpikes
+            dataKey="DataCalibrated"
+            outputKey="DataCalibrated"
+            startCellKey="StartCellData"
+            leftBorder="${RemoveSpikes.leftBorder}"
+            spikeLimit="${RemoveSpikes.spikeLimit}"
+            topSlopeLimit="${RemoveSpikes.topSlopeLimit}"
+            maxSpikeLength="${RemoveSpikes.maxSpikeLength}"
+        />
+        <fact.datacorrection.InterpolateTimeSeries
```
jebuss: Not sure if we should interpolate the bad pixels here, or if they should rather be marked and interpolated in the analysis

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,114 @@
...
+                maxSpikeLength="${RemoveSpikes.maxSpikeLength}"
+        />
+
+        <fact.datacorrection.DrsTimeCalibration
+                startCellKey = "LONS_StartCellData"
+                dataKey      = "LONS_DataCalibrated"
+                outputKey    = "LONS_DataCalibrated"
+        />
+
+        <fact.datacorrection.InterpolateTimeSeries
```
jebuss: same as above

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,114 @@
...
+        />
+        <fact.utils.Remapping
+                key="Data"
+                outputKey="Data"
+                reverse="True"
+        />
+        <!--<fact.ShowViewer key="DataCalibrated" />-->
+
+        <fact.io.FITSWriter url="${output}"
+                            keys="${OUTPUT_KEYS}"
```
jebuss: could this rather be realised with wildcards and exclusions?

examples/pedestal_noise_mc_analysis.xml
```diff
@@ -0,0 +1,75 @@
+<container>
```
jebuss: could you add a doc string that is telling what the process' purpose and what it is doing?

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,114 @@
...
+                            headerKeys="CREATOR,NROI,NPIX,TELESCOP,RUNTYPE,CAMERA"
```
jebuss: should the value of CREATOR be changed to indicate how the MCs were created e.g. Ceres_<svn revision number>_FACT++_<svn revision> or something like that

examples/pedestal_noise_mc_analysis.xml
jebuss: BTW why is this process needed?

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+import static java.lang.Math.ceil;
+
+/**
+ * This processors changes the order of the pixels in the data from SoftId to Chid
```
jebuss: Lying documentation

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+ *
+ * @author mbulinski
+ *
+ */
+public class SamplePedestalEvent implements StatefulProcessor {
+    static Logger log = LoggerFactory.getLogger(SamplePedestalEvent.class);
+
+    @Parameter(required = false, defaultValue="0", description = "The seed to make the results reproductable if desired.")
+    private long seed;
```
jebuss: it says default = 0, but there is no default

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+
+    @Parameter(required = true, description = "The file containing the noise database for the noise to be used.")
+    private SourceURL noiseDatabase;
+
+    @Parameter(required = true, description = "The folder containing the data files.")
+    private String dataFolder;
```
jebuss: why isn't this a Source URL?

jebuss: add to description: " in FACTS canonical folder structure."

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+        item.put("drspath", new File(drsPath));
+        item.put("prevEvents", previousEventInfo);
+        return item;
+    }
+
+    /**
+     * load the database and apply the condition and binning
+     */
+    private void prepareNoiseDatabase() {
+        this.database = new ArrayList<Data>(100000);
```
jebuss: Magic number, add comment why

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+
+    @Parameter(required = false, description = "The condition for the noise which will be used")
+    private String noiseCondition ="";
+    private Condition condition;
+
+    @Parameter(required = true, description="The binning of the zenith-angle. If single value: steplenght, bins into the given steplength"+
+                                            "If multivalue: Each value is the length of a single step starting with zenith=0.")
+    private String[] binning;
+
+    @Parameter(required = true, description = "The binning key of the double value from the noise database.")
```
jebuss: is this the key of the Zd value in the noise DB?

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+    private String dbBinningKey;
+
+    @Parameter(required = true, description = "The binning key of the double value from the item.")
```
jebuss: is this the key of the Zd value in the file of the event?

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+        ConditionFactory factory = new ConditionFactory();
+        this.condition = factory.create(noiseCondition);
+    }
+
+
+    public void setBinning(String[] binning) {
+        this.binning = binning;
+        if (binning.length==1) { // single value
+            double steplength = Double.parseDouble(this.binning[0]);
+            int numBins = (int)ceil(90.0/steplength);
```
jebuss: could this whole code in the setter be encapsulated in a meaningful function and have a unit test?

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    /**
+     * load the database and apply the condition and binning
+     */
+    private void prepareNoiseDatabase() {
+        this.database = new ArrayList<Data>(100000);
+        this.bins_index = new ArrayList<List<Integer>>(this.bins.length);
+        for(int i=0; i<this.bins.length; i++) {
+            this.bins_index.add(new ArrayList<Integer>(100));
```
jebuss: annother Magic number

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+        }
+        try {
+	        log.info("Noise db: "+noiseDatabase.toString());
+            JSONStream databaseStream = new JSONStream(noiseDatabase);
+            databaseStream.init();
+
+            for(Data item = databaseStream.readNext(); item!=null; item = databaseStream.readNext()) {
```
jebuss: wouldn't a while loop fit better?

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+
+    /**
+     * Given a value data, calculate the bin it would end up in.
+     *
+     * @param data
+     * @return
+     */
+    private int getBin(double data) {
+        int i=0;
+        while(i<this.bins.length) {
```
jebuss: why here no for loop?

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+            return fullpath+".fits.gz";
+        }
+        throw new RuntimeException("Couldn't find data file: "+fullpath+".fits.*");
+    }
+
+    /**
+     * Creates the path to the drs file and checks whether it exists
+     * @param startFolder
+     * @param night
+     * @param drs0
```
jebuss: what are drs0 and drs1

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+        } catch (Exception e) {
+            log.error("Couldn't read file: "+fullpath);
+            throw new RuntimeException(e);
+        }
+
+        if (item==null) {
+            log.error("Couldn't read the pedestal item, nr: '"+noiseNr+"' file: "+fullpath);
+            throw new IOException("Couldn't read the pedestal item");
+        }
+        item.put("noiseNr", noiseNr);
```
jebuss: I'd like to have a more descriptive name instead of noiseNr. I guess it is the event ID of the sampled pedestal event. 
So how about pedestal_event_id, or background_event_id or so

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,350 @@
...
+        // test .gz
+        noiseFile = new File(fullpath+".fits.gz");
+        if (noiseFile.exists()) {
+            return fullpath+".fits.gz";
+        }
+        throw new RuntimeException("Couldn't find data file: "+fullpath+".fits.*");
+    }
+
+    /**
+     * Creates the path to the drs file and checks whether it exists
```
jebuss: This need explaination how the DRS file ist selected. What does it look for, the temporal closest, the last file?

examples/add_pedestal_noise_template.xml
tarrox: Done.

examples/add_pedestal_noise_template.xml
tarrox: Would be nice but there is no good way to exclude all keys that are not needed, we would need a better way of naming our keys first (for example all header keys should for example start the same way).

examples/add_pedestal_noise_template.xml
tarrox: Done changed to Ceres_Pedestal.

examples/add_pedestal_noise_template.xml
tarrox: You do not need to mark anything you would just need to take the 'LONS_UnixTimeUTC' and use it to calibrate it in the 'InterpolateTimeSeries'. Actually this would make sense as we only want to calibrate with the error pixel on that day. So we could remove this here theoretically and change the analysis side of things.

examples/add_pedestal_noise_template.xml
tarrox: read above

examples/pedestal_noise_mc_analysis.xml
tarrox: Right now it is due to some errors.

jebuss:
I added a recursive  call of the function that is sampling the pedestal event. This way we should sample a new event in case reading the first fails. see `SamplePedestalEvent.samplePedestalEventByBin`

examples/add_pedestal_noise_template.xml
jebuss: yes, I think this should be done. I think this case should be handled in https://github.com/fact-project/fact-tools/blob/2098a0fdcf4f0d3fae17e1049db1b947aead0cc9/src/main/resources/analysis/calibration.xml#L40. I think depending on the 'Creator' a different 'timeStampKey' should be used in https://github.com/fact-project/fact-tools/blob/2098a0fdcf4f0d3fae17e1049db1b947aead0cc9/src/main/resources/analysis/calibration.xml

examples/pedestal_noise_mc_analysis.xml
jebuss: anyways, doc string please

src/main/java/fact/utils/SamplePedestalEvent.java
jebuss: please add to description: "e.g. Zd value in the provided noise DB"

src/main/java/fact/utils/SamplePedestalEvent.java
jebuss: please add to description: "e.g. key of the zd angle of the provided mc event. The value will be compared to the value of the dbBinningKey "

src/main/java/fact/utils/SamplePedestalEvent.java
jebuss: In that case: Please add an explanation what `drs0` and `drs1` are, and by which priority they are chosen for the drs calibration of the event

tarrox:
Added more comments and changed the multiple try code a little bit. Now it is also adding the amount of tries the sampling processor took for each event and writes them to the resulting file. Generally we want the average of this column to be as close to 0 (only one try for each event, if its not, stuff needs fixing badly).

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,130 @@
...
+        @parameters SET_dataFolder The folder containing the fact files, structured in the fact data folder convention.
+        @parameters input          The multidata json file (see processor). Containing the MC data which should be added.
+        @output     output         The name of the finished fits file.
+    -->
+    <properties url="classpath:/default/settings.properties" />
+
+    <service id="calibService" class="fact.calibrationservice.ConstantCalibService" />
+    
+    
+    <property name="noiseDB" value="file:${SET_noiseDB}" />
```
MaxNoe: Why SET? Why inconsistent use of capitalisation, snake and camel case?

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,130 @@
...
+                outputKey="Data"
+                reverse="true"
+        />
+        <fact.utils.Remapping
+                key="Data"
+                outputKey="Data"
+                reverse="True"
+        />
+        
+        <stream.script.JavaScript>
```
MaxNoe: Don't we maybe want to set the creator in the processor that actually applies the noise instead of using inline javascript in the xml?

src/main/java/fact/io/hdureader/FITS.java
```diff
@@ -83,6 +83,7 @@ public static FITS fromFile(File file) {
     public FITS(URL url) throws IOException {
         this.url = url;
         DataInputStream stream = getUnGzippedDataStream(url);
+        log.info("Reading file: "+url.toString());
```
MaxNoe: info? or debug?

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,385 @@
...
+import static java.lang.Math.ceil;
+
+/**
+ * This processors loads the given noise database and randomly chooses for each event processed a noise event,
+ * which is then loaded into the item, with the keys prepended with the string prependKey.
+ *
+ * @author mbulinski
+ *
+ */
+public class SamplePedestalEvent implements StatefulProcessor {
```
MaxNoe: We made all parameters public in the other processors.

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,385 @@
...
+
+    protected int getBin(double data) {
+        for(int i=0; i<this.bins.length; i++) {
+            if (data<this.bins[i])
+                return i;
+        }
+        throw new RuntimeException("Data has a value outside of the binning");
+    }
+
+    public void setNoiseFolder(String dataFolder) {
```
MaxNoe: Why are getters and setters needed?

src/main/java/fact/io/hdureader/FITS.java
tarrox: This should be debug and will be in the next commit.

src/main/java/fact/utils/SamplePedestalEvent.java
tarrox: Fixed.

src/main/java/fact/utils/SamplePedestalEvent.java
tarrox: Historically still in there, but are now gone.

examples/add_pedestal_noise_template.xml
tarrox: The apply doesn't really happens until latter in the `CombineDataArrays` and not in the `SamplePedestalEvent` processor. Or to put it more simply the whole set of processors create the finished product, that is why it is set here explicitly.

We could also say that the 'SamplePedestalEvent' processor changes the creator but on its own it just adds additional stuff to the data item and doesn't change anything in the original item.

examples/add_pedestal_noise_template.xml
tarrox: In this case it is just written like this, because this value needs to be set explicitly beforehand as erna only sets input, output and aux_source. It basically tells the user to SET the `noiseDB` parameter. Which he can do manually by hand or more automatically through a text processor.

examples/add_pedestal_noise_template.xml
MaxNoe: > The apply doesn't really happens until latter in the CombineDataArrays and not in the SamplePedestalEvent processor. Or to put it more simply the whole set of processors create the finished product, that is why it is set here explicitly.

Is there any usecase of using just one of the processors? If not, maybe we should incroparate this into only one processor.

examples/add_pedestal_noise_template.xml
MaxNoe: I don't get it. So we have a property that get's set to the value of `SET_` plus that name. Why? Why not directlty set the property?

examples/add_pedestal_noise_template.xml
tarrox: The problem with applying it into a single processor is, that the two processors are not side by side. Because we first need to calibrate the event before we can combine them, so there are other processor in between and putting all of that into a single processor is not the nicest thing to do.

I think the best solution will be to export the whole thing into an module that can kinda act like a single processor, with the parameters set trough properties (I don't think modules have parameters, right?). That way users can use it as a kinda processor but we don't get all the additional code if we make a mega processor (It also allows for easy changes to the calibration part and so on). Because if you think about it this whole thing is like a processor made out of processors.

jebuss: Yes I agree to Michael putting everything in one processor creates a lot copy code (e.g. Drs_Calibrations) that we should avoid. Thus I up for keeping the process in several containers and would vote for adding a module like the cleaning.xml

src/main/resources/analysis/init.xml
```diff
@@ -1,11 +1,14 @@
...
-  <If condition="%{data.CREATOR}!='Ceres'">
+  <If condition="%{data.CREATOR}!='Ceres' and %{data.CREATOR}!='Ceres_Pedestal'">
     <fact.utils.PreviousEventInfo
         startCellKey="StartCellData"
         outputKey="prevEvents"
         limitEvents="${PreviousEventInfo.limitEvents}"
     />
     <fact.features.UnixTimeUTC2DateTime />
   </If>
+  <If condition="%{data.CREATOR}=='Ceres_Pedestal'">
```
jebuss: is this still needed. If I am not mistaken all processors are using ZonedDateTime now and don't need a in time stamp anymore.

examples/add_pedestal_noise_template.xml
MaxNoe: If that's the case, just leave it empty and do not imply there is a `SET_<name>` property.

src/main/java/fact/features/source/SourcePosition.java
```diff
@@ -136,8 +136,7 @@ public Data process(Data item) {
 
         } else {
             // Assume observations
-            Utils.isKeyValid(item, timeStampKey, ZonedDateTime.class);
-            ZonedDateTime timeStamp = (ZonedDateTime) item.get(timeStampKey);
+            ZonedDateTime timeStamp = Utils.getTimeStamp(item, timeStampKey);
```
MaxNoe: I think the fallback solution of `Utils.getTimeStamp` is not wanted here

src/main/java/fact/utils/SamplePedestalEvent.java
MaxNoe: The fields are all stil private

src/main/java/fact/utils/SamplePedestalEvent.java
MaxNoe: No, they are still here

examples/add_pedestal_noise_template.xml
jebuss: placet

examples/add_pedestal_noise_template.xml
jebuss: @tarrox could we easily create a noiseDB file for the events in the test data in the rescources? Then one could add this to the repo and also include it in the tests.

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,130 @@
...
+    
+    <property name="OUTPUT_KEYS" value="${MC_KEYS},${LONS_KEYS}"/>
+
+    <stream id="fact" class="fact.io.FactFileListMultiStream"   url="${input}">
+        <stream class="fact.io.hdureader.FITSStream" id="fact" />
```
jebuss:
```suggestion
        <stream class="fact.io.CeresStream" id="fact" />
```

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,130 @@
...
+
```
jebuss:
```suggestion
    <property name="MC_HEADER_KEYS"         value="MCorsikaRunHeader.*,RUNID,MMcRunHeader.*,MCorsikaRunHeader.*,MMcCorsikaRunHeader.*,MMcRunHeader.*,RUNTYPE"/>
    <property name="MC_KEYS"   value="IncidentAngle.fVal,MCorsikaEvtHeader.*,CorsikaEvtHeader.fFirstInteractionHeight,MMcEvt.*,MMcEvtBasic.*,MPointingPos.*,MRawEvtData.fHiGainPixId,Data,StartCellData,TriggerType,MRawEvtData.*,EventNum,MRawEvtHeader.*,MSimSourcePos.*,McCherPhot*,McMuonCherPhot*,McNoisePhot*,McCherArrTime*,"/>
```

examples/add_pedestal_noise_template.xml
```diff
@@ -0,0 +1,130 @@
...
+    
+    <property name="OUTPUT_KEYS" value="${MC_KEYS},${LONS_KEYS}"/>
```
jebuss:
```suggestion
    <property name="OUTPUT_KEYS" value="${MC_HEADER_KEYS},${MC_KEYS},${LONS_KEYS}"/>
```

examples/add_pedestal_noise_template.xml
jebuss: added wildcards above

src/main/java/fact/utils/SamplePedestalEvent.java
jebuss: Made this into an issue. Not crucial for this PR

src/main/java/fact/utils/SamplePedestalEvent.java
```diff
@@ -0,0 +1,385 @@
...
+
+    @Parameter(required = false, description = "The condition for the noise which will be used")
+    private String noiseCondition ="";
+    private Condition condition;
+
+    @Parameter(required = true, description="The binning of the zenith-angle. If single value: steplenght, bins into the given steplength"+
+                                            "If multivalue: Each value is the length of a single step starting with zenith=0.")
+    private String[] binning;
+
+    @Parameter(required = true, description = "The binning key of the double value from the noise database.")
```
jebuss:
```suggestion
    @Parameter(required = true, description = "The binning key of the double value from the noise database. e.g. key of the Zd value in the provided noise DB. A bin will be chosen according to the itemBinningKey")
```

examples/pedestal_noise_mc_analysis.xml
```diff
@@ -0,0 +1,48 @@
...
+  <process id="2" input="fact">
+    <include url="classpath:/analysis/init.xml" />
+    <!-- This skips events created by our malfunctioning LED light pulser (27s Events) --> 
+    <Skip condition="%{data.saturated_slices_percentage} &gt; 0.2" />
+    <!-- Only analyze physics trigger events -->
+    <Skip condition="%{data.TriggerType} != 4" />
+
+    <include url="classpath:/analysis/calibration.xml" />
+    <include url="classpath:/analysis/extraction.xml" />
+    <include url="classpath:/analysis/cleaning.xml" />
```
jebuss: you took off the adapted cleaning step for the `LONS_timestamp`.  If I am not misstaken, the cleaning will now use the timestamp of the MC file which will lead to a wrong calculation of dead/crazy pixels and the postion of Stars in the FOV. Maybe it makes sense to overwrite the value of the time stamp with `LONS_timestamp`?

mackaiver:
Looks very smakelijk.

PR #374, opened 2018-10-19
Add metadata to both fits headers

MaxNoe:


src/main/java/fact/io/FITSWriter.java
```diff
@@ -289,6 +282,19 @@ private void initFITSFile(Data item, Data headerItem) throws FitsException {
 
     }
 
+    public void addFACTToolsData(Header header) throws FitsException{
```
mackaiver:
```suggestion
    public void addFACTToolsMetaData(Header header) throws FitsException{
```

src/main/java/fact/io/FITSWriter.java
```diff
@@ -298,11 +304,7 @@ public void init(ProcessContext processContext) throws Exception {
 
         // We first have to write an empty header because a binary table cannot be the first hdu
         BasicHDU bhdu = BasicHDU.getDummyHDU();
-        bhdu.getHeader().addValue(
-                "VERSION",
-                VersionInformation.getInstance().gitDescribe,
-                "The FACT-Tools Version used to write this file"
-        );
+        addFACTToolsData(bhdu.getHeader());
```
mackaiver:
```suggestion
        addFACTToolsMetaData(bhdu.getHeader());
```

src/main/java/fact/io/FITSWriter.java
```diff
@@ -289,6 +282,19 @@ private void initFITSFile(Data item, Data headerItem) throws FitsException {
...
 
+    public void addFACTToolsData(Header header) throws FitsException{
+        header.addValue(
+                "VERSION",
+                VersionInformation.getInstance().gitDescribe,
+                "The FACT-Tools Version used to write this file"
+        );
+        header.addValue("ORIGIN", "FACT", "");
+        header.addValue("TELESCOP", "FACT", "");
+        header.addValue("INSTRUME", "FACT", "");
```
jebuss: Stupid question: Missing "NT" or not more than 8 character possible?

PR #388, opened 2018-12-15
Gain rescaling

jebuss:
WIP:

adds a processor to rescale a data array from mc gain to data gain and vice versa. This is necessary for the pedestal superposition since MC and data gain differ intrinsically. Accordingly the MCs have to be calibrated with the MC gain and the pedestal events with the data gain.

In order to overcome this, this processor allows to rescale the data array (meaning the time series) from mc gain to data gain, by dividing the time series of each pixel through the pixels data gain and multiplying with the mc gain.

### DIFF
--- a/src/main/java/fact/pedestalSuperposition/rescaleDataArray.java
+++ b/src/main/java/fact/pedestalSuperposition/rescaleDataArray.java
@@ -1,0 +1,113 @@
+package fact.pedestalSuperposition;
+
+import fact.Constants;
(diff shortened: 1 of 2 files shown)


jebuss:
This was an awesome idea, but it is bullshit. It does not make things better. Rather the opposite, since it shifts the pedestal features. e.g.
ped_var_mean

I know that the difference of the data and mc game is missing and not considered  but this does not solve it. Thus, I close the PR.

Issue #389, opened 2018-12-17
UnixTimeUTC and BoardTime are repeated for 100 times


MaxNoe:
When reading zfits files, 100 consecutive events have the same `UnixTimeUTC` and `BoardTime`.

This smells like a bug related to zfits tiles.

MaxNoe:
From https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#asShortBuffer

> The content of the new buffer will start at this buffer's current position. Changes to this buffer's content will be visible in the new buffer, and vice versa; the two buffers' position, limit, and mark values will be independent. 

So the buffer does not get advanced.

MaxNoe:
fixed in #390

MaxNoe:
Now I came to see that this also affected `StartCellData`, so all our DRS calibrations also must have been wrong since that change was made.

MaxNoe:
So this should affect all files with `ZTILELEN` > 1

dneise:
> Now I came to see that this also affected StartCellData, so all our DRS calibrations also must have been wrong since that change was made.

Was this not realized, since the data in the fits files are already DRS-amplitude calibrated a little bit, and then applying a wrong DRS-calibration is not so easily visible anymore?

But still. Does this mean, the currently running analysis should better be restarted?

MaxNoe:
>      Was this not realized, since the data in the fits files are already DRS-amplitude calibrated a little bit, and then applying a wrong DRS-calibration is not so easily visible anymore? 
But still. Does this mean, the currently running analysis should better be restarted?

I jsut checked the `ZTILELEN` of the files at ISDC, only files from 2013 until mid October have `ZTILELEN == 100`, after that, most have `ZTILELEN == 1`, a few have small integer numbers.

This is probably  why it never was realized. Only files before 2013-10-22 are affected heavily. All other files are `ZTILELEN == 1, 2, 3`, so not or only not as badly affected.

PR #390, opened 2018-12-17
Fix zfits boardtime

MaxNoe:


src/main/java/fact/io/hdureader/ZFITSHeapReader.java
```diff
@@ -674,43 +674,52 @@ private Serializable readSingleValueFromBuffer(BinTable.TableColumn c, ByteBuffe
...
+            case CHAR:
+                char[] chars = new char[c.repeatCount];
+                for (int i = 0; i < c.repeatCount; i++) {
+                    chars[i] = buffer.getChar();
+                }
+                return chars;
             case SHORT:
                 short[] shorts = new short[c.repeatCount];
-                buffer.asShortBuffer().get(shorts);
+                for (int i = 0; i < c.repeatCount; i++) {
```
mackaiver: does this have a performance impact?

MaxNoe:
What's the Performance impact if it was wrong before?

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
MaxNoe: What's the Performance impact if it was wrong before?

dneise:
Is the fact that the buffer was not advanced a bug in get() ?

MaxNoe:
No, as documented in #389, the calls to `asXXXXBuffer` create a new view, that by design has an independent position. So this is the intended behaviour. We just used it wrong.

dneise:
Ah I missed that link to the documentation .. so this was wrong all the time and in all the places you fixed it ... and it was never realized. ... that is ... somehow depressing.

src/main/java/fact/io/hdureader/ZFITSHeapReader.java
mackaiver: About a factor of 6 if I remember correctly. Also to my understanding the `buffer.asShortBuffer().get(shorts);` should be equivalent to the loop. I don't understand where this goes wrong.

mackaiver:
> No, as documented in #389, the calls to `asXXXXBuffer` create a new view, that by design has an independent position. So this is the intended behaviour. We just used it wrong.

Aha. That's surprising.

mackaiver:
Can we advance the buffer by hand? We replaced all those loops a while ago on purpose.

MaxNoe:
Replaced the loops with reading into arrays by directly putting the correct buffer views into the tile cache.

PR #391, opened 2019-10-11
Better error message for empty FITS files

MaxNoe:
, allow only skipping empty files in FileListMultiStream


MaxNoe:
This changes an IndexError to `MissingHDUError` with message `File did not contain Events HDU`

MaxNoe:
Okay, but it does not harm in 3.6, just a 3 line function doing nothing.

However, as I found out that python normalizes all identifiers to NFKC, the problem is much smaller than I thought.

So just adding three characters for Angstrom,  Ohm and Micro would be all, the degree sign is not a valid identifier, neither are the super scripts.

